PR #27, opened 2019-06-07
Bug fix/accordion arrow

### LINKED ISSUE #23: Accordion: Arrow direction bug
**Describe the bug**
The direction of the arrow if wrong.

**Expected behavior**
Should face down when closed and up when expanded.

**Additional context**
The original specification was correct. I just made an error.

mario-subo:
fixes #23


src/Accordion/Accordion.tsx
```diff
@@ -46,7 +46,7 @@ export class Accordion extends React.Component<AccordionProps, AccordionState> {
                 {this.props.list.map((item: AccrodionListItem, index: number) =>
```
yousifalraheem: this will crash when the `list` prop is accidentally set to `null`. Use this instead `this.props.list && this.props.list.map`

mario-subo: resolved in commit https://github.com/sebgroup/react-components/pull/27/commits/82e4778e78a4c6f887280eb5c2db59b485b9c99f

PR #31, opened 2019-06-11
Refactor/Default Html Attributes

### LINKED ISSUE #25: Include default HTML attributes as component props
I wanted to add an `id` attribute to `Button` component but am unable to do so. It is requested by QA so they can target specific buttons when writing Selenium tests.

Ability to pass relevant HTML attributes as props, similar to [Material UI](https://github.com/mui-org/material-ui)

1. Wrapping `Button` in a div and use `$("#button-wrapper > button)` selector instead, but it's additional effort on both development and QA's side
2. Using classes instead for targeting. This is not always viable. Certain business cases require more specific targeting

mario-subo:
adds the ability to provide and id and name to the underlying html element for some of the
components which were missing this feature before

fix #25 re #25


src/CheckBox/CheckBox.tsx
```diff
@@ -25,7 +26,7 @@ export const CheckBox: React.FunctionComponent<CheckBoxProps> = (props: CheckBox
                     <input
                         type="checkbox"
                         className="custom-control-input"
-                        id={props.name}
+                        id={props.id || props.name}
```
yousifalraheem: Id should be random if not set. Name and ID should be separated.

mario-subo: resolved in commit https://github.com/sebgroup/react-components/pull/31/commits/cd1e9c4a4dbca52693e196eeec05a0303300e026

PR #34, opened 2019-06-18
Refactor/adding id props

### LINKED ISSUE #25: Include default HTML attributes as component props

### LINKED ISSUE #45: Add support for pattern in textbox
### Description
Add support for **pattern** in TextBox and TextBoxGroup components. The reason is that if I wanted to trigger number pad keyboard on iOS I need to pass patter `\d*`

mario-subo:
this fixes #25 fixes #45


DanSallau:
You didn't add for toggle, textBoxGroup, textBox, radio etc. @mario-subo

PR #36, opened 2019-06-21
Feature/accessability

DanSallau:


src/RadioButton/radio-button-style.scss
```diff
@@ -28,6 +28,9 @@ $red: #F03529;
                 &:hover+span {
                     border-color: $blue;
                 }
+                &:focus+span {
+                    outline: 1px solid #A5C7FE;
```
omarVodiaK: is it possible to use a variable here ?

src/Tabs/Tabs.tsx
```diff
@@ -13,24 +13,103 @@ interface TabsProps {
...
+        if (this.props.onClick && !item.disabled) {
+            console.log("Its been called ", item);
+            this.props.onClick(i);
+        }
+    }
+
+    /**
+     * e: Event
+     */
+    onKeyDown(e: React.KeyboardEvent<HTMLAnchorElement>) {
```
mario-subo: Since you're repeating `e.key.toLowerCase()` and `this.props.list` throughout the scope of this function I suggest for better readability to refactor like so:

```
const key = e.key.toLowerCase();
const { list } = this.props;
const { localActiveTab } = this.state;

if (localActiveTab < list.length && localActiveTab >= 0) {
...
}

 if ((key === "arrowleft" || key === "arrowdown") {
...
}


```

PR #43, opened 2019-08-27
feat(modal-component): new modal component

omarVodiaK:
This Modal is similar to the dialogue component.
The modal is able to accept HTML header, body and footer.
The modal is able to be toggled in full screen, to the right and left.
**Normal Modal Desktop:**
Modal
**Modal Left:**
Desktop Modal left
**Modal Right:**
Desktop Modal Right
**Modal Fullscreen:**
Desktop Full Screen
**Normal Modal Mobile:**
Mobile Modal
**Mobile Fullscreen:
Mobile fullscreen
**


develop/components/pages/ModalPage.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from "react";
+import { Modal } from "../../../src/Modal/Modal";
+import { Button } from "../../../src/Button/Button";
+import { getParameterByName } from "../../utils/queryString";
+const Highlight = (require("react-highlight")).default;
+const docMD = require("../../../src/Modal/readme.md");
+
+export default class ModalPage extends React.Component<any, any>  {
```
DanSallau: I noticed we have a state of type any for all the component pages despite the fact that have states in all the components. I recommend we should change all of them starting with this, because this project being open source means the code quality is open to criticism.

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <div
+                className={"modal-backdrop" + (this.state.open ? " show" : " fade")}
+                onClick={(event: React.MouseEvent) => { !this.props.disableBackdropDismiss && this.closeModal(event); }}
+            >
+                <div role="dialog" tabIndex={-1} className={"modal " + this.setModalClasses()}>
+                    <div role="document" className="modal-dialog" tabIndex={-1}>
+                        <div className="modal-content">
+                            <div className="modal-header">{this.props.header}</div>
+                            <div className="modal-body">{this.props.body}</div>
+                            <div className="modal-footer">{this.props.footer}</div>
```
yousifalraheem: Use conditional rendering since **header**, **body** and **footer** are optional props
```jsx
{this.props.header && <div className="modal-header">{this.props.header}</div>}
{this.props.body && <div className="modal-body">{this.props.body}</div>}
{this.props.footer && <div className="modal-footer">{this.props.footer}</div>}
```

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ariaDescribedby?: string;
+    dismissModal?: () => void;
+    header?: React.ReactNode;
+    body?: React.ReactNode;
+    footer?: React.ReactNode;
+}
+
+interface ModalState {
+    open: boolean;
+    close: boolean;
```
yousifalraheem: this state looks useless, maybe use one state `toggle`

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface ModalProps {
+    toggle: boolean;
+    id?: string;
+    fullscreen?: boolean;
+    position?: "left" | "right";
+    className?: string;
+    disableBackdropDismiss?: boolean;
+    ariaLabel?: string;
+    ariaDescribedby?: string;
+    dismissModal?: () => void;
```
yousifalraheem: use JavaScript naming convention for methods `onDismiss`. Working with React it's just easy to remember without a manual. All methods are named similarly (e.g. `onClick`, `onHover`)

src/Modal/Modal.tsx
yousifalraheem: I think you can rely on the `toggle` prop instead of having a state. I suggest converting this component to `React.FunctionComponent` since removing both states would defeat the purpose of using `React.Component`

this process would remove the need for `constructor`, `componentDidMount` and `componentDidUpdate`

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            .map(([key, value]) => key)
+            .join(" ");
+    }
+
+    /**
+     * emit dismissModal event when backdrop is clicked
+     * @param {React.MouseEvent} event clicked element
+     */
+    closeModal(event: React.MouseEvent): void {
+        const target: HTMLElement = event.target as HTMLElement;
```
yousifalraheem: you don't need this, just declare `event: React.MouseEvent<HTMLDivElement>` and you should have access to target.

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+        if (event && event.target && target.classList && target.classList.length) {
+            const classList: DOMTokenList = target.classList;
+            if (classList.contains("modal")) {
+                this.setState({ open: false, close: true });
+                this.props.dismissModal();
```
yousifalraheem: `dismissModal` is declared as an optional prop and it would cause the app to crash in this case. I would suggest making it required instead. Also check before executing `this.props.dismissModal && this.props.dismissModal();`

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+
+    render() {
+        return (
+            <div
+                className={"modal-backdrop" + (this.state.open ? " show" : " fade")}
+                onClick={(event: React.MouseEvent) => { !this.props.disableBackdropDismiss && this.closeModal(event); }}
```
yousifalraheem: no need to put the logic here in the template, just use this instead (with renaming it):
```jsx
onClick={this.onDismiss}
```
make sure to bind the method in the `constructor`
```typescript
this.onDismiss = this.onDismiss.bind(this);
```
and move the logic to the `onDismiss` method

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+        return classes;
+    }
+
+    /**
+     * Convert object to string
+     * @param { [key: string]: boolean } classes list of classes to be appended
+     * @returns {string} string of concatenated class names
+     */
+    classNames(classes: { [key: string]: boolean }): string {
+        return Object.entries(classes)
```
yousifalraheem: `Object.entries` is [not supported in IE](https://caniuse.com/#search=Object.entries) use `Object.keys` instead
```typescript
return Object.keys(classes).filter((key: string) => classes[key]).join(" ");
```

src/Modal/Modal.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setModalClasses(): string {
+        const classes: string = this.classNames({
+            "show": this.state.open,
+            "fade": this.state.close,
+            "modal-aside": !!this.props.position,
+            "modal-aside-left": this.props.position === "left",
+            "modal-aside-right": this.props.position === "right",
+            "modal-fullscreen": this.props.fullscreen,
+            [this.props.className]: !!this.props.className
+        });
```
yousifalraheem: I don't know why you chose this to be a two-step process? Declaring an object and then processing this object. You could do this instead:
```typescript
let classes: string = this.props.toggle ? "show" : "hide";
classes += !!this.props.position ? " modal-aside" : "";
classes += !!this.props.position === "left" ? " modal-aside-left" : " modal-aside-right";
classes += this.props.fullscreen ? " modal-fullscreen" : "";
classes += this.props.className ? ` ${this.props.className}` : "";
return classes;
```
I think you're just using `ngClass` logic from `Angular` here and adding a method to process it which is unnecessary.

src/Modal/Modal.tsx
omarVodiaK: onDismiss is a prop it shouldn't trigger if the user choose to disable the back drop dismiss. So we need to keep the condition
and the event should not trigger only when the user clicks in the backdrop, your suggestion will not handle those cases.

src/Modal/Modal.tsx
omarVodiaK: i believe its a cleaner solution than concatenating string and adding spaces to class names.

src/Modal/Modal.tsx
omarVodiaK: `const target: HTMLElement = event.target as HTMLElement ` will solve the issue with line 80 `target.classList`

develop/components/pages/ModalPage.tsx
omarVodiaK: I dont think we can use `ModalProps` for ModalPage, 
if you check line 29 contains `this.props.location` that has nothing to do with `ModalProps`
I also dont believe that props of modal and props of a page are similar.

DanSallau: No @omarVodiaK i am not saying they are similar and also what i am trying to point out is nit just peculiar to this PR. I am suggesting we should perhaps to it do the remaining page so here

```
interface ModalPageState {
  dialogue: boolean;
  fullscreen: boolean;
  position: string;
  disableBackdropDismiss: boolean;
} 
```

and 

export default class ModalPage extends React.Component<any, ModalPageState >  {

src/Modal/Modal.tsx
yousifalraheem: I'm trying to avoid casting the `event.target as HTMLElement` since we don't know why update in the future might we get that changes that type since these events are **[React synthetic events](https://reactjs.org/docs/events.html)**. In order to access the properties of the target, you use this `event.currentTarget.classList` which will inherit the properties of the element you pass to the type constructor which in this case `React.MouseEvent<HTMLDivElement>` takes the properties of the `Div` element.

src/Modal/Modal.tsx
yousifalraheem: The method that I'm suggesting `this.onDismiss` is different from `this.props.onDismiss` that's why I was asking you to bind it in the constructor since it's a local method that hosts whatever logic is related to dismissing the modal. You already have such a method, it's called `closeModal`. So my previous comment was to suggest that you rename the method and move this logic inside it. You could keep the same name but at least change the `JSDoc` since it says that it will **emit** a dismiss modal, I guess this comment was copied directly from its Angular counterpart.

PR #50, opened 2019-10-01
Feature/accordion alternative style

kherP:
added new property `alternative` to allow to display accordion in alternative style. 

image

image

### DIFF
--- a/src/Accordion/Accordion.test.tsx
+++ b/src/Accordion/Accordion.test.tsx
@@ -130,13 +130,8 @@ describe("Component: Accordion", () => {
         expect(mountedWrapper.find(".text-wrapper").length).toBeFalsy();
     });
 
-    describe("Should render alternative theme", () => {
-        beforeEach(() => {
-            mountedWrapper = mount(<Accordion list={accordionList} alternative={true} />);
-        });
-
-        it("Should render alternative theme", () => {
-            expect(mountedWrapper.find(".custom-accordion").hasClass("alternative-accordion")).toBeTruthy();
-        });
+    it("Should render alternative theme", () => {
+        mountedWrapper = mount(<Accordion list={accordionList} alternative={true} />);
+        expect(mountedWrapper.find(".custom-accordion").hasClass("alternative-accordion")).toBeTruthy();
     });
 });
--- a/src/Accordion/Accordion.test.tsx
+++ b/src/Accordion/Accordion.test.tsx
@@ -129,4 +129,10 @@ describe("Component: Accordion", () => {
         mountedWrapper.setProps({ list: [] });
         expect(mountedWrapper.find(".text-wrapper").length).toBeFalsy();
     });
+
+    it("Should render alternate theme", () => {
+        const alternative: boolean = true;
+        wrapper.setProps({ alternative });
+        expect(wrapper.hasClass("alternate-accordion")).toBeTruthy();
+    });
 });
--- a/src/Accordion/Accordion.test.tsx
+++ b/src/Accordion/Accordion.test.tsx
@@ -130,13 +130,13 @@ describe("Component: Accordion", () => {
         expect(mountedWrapper.find(".text-wrapper").length).toBeFalsy();
     });
 
-    describe("Should render alternate theme", () => {
+    describe("Should render alternative theme", () => {
         beforeEach(() => {
             mountedWrapper = mount(<Accordion list={accordionList} alternative={true} />);
         });
 
-        it("Should render alternate theme", () => {
-            expect(mountedWrapper.find(".custom-accordion").hasClass("alternate-accordion")).toBeTruthy();
+        it("Should render alternative theme", () => {
+            expect(mountedWrapper.find(".custom-accordion").hasClass("alternative-accordion")).toBeTruthy();
         });
     });
 });
--- a/src/Accordion/Accordion.test.tsx
+++ b/src/Accordion/Accordion.test.tsx
@@ -25,9 +25,9 @@ describe("Component: Accordion", () => {
     it("Should render custom className and id", () => {
         const className: string = "myAccordionClass";
         const id: string = "myAccordionId";
-        wrapper.setProps({ className, id });
-        expect(wrapper.hasClass(className)).toBeTruthy();
-        expect(wrapper.find(`#${id}`).length).toBeGreaterThan(0);
+        mountedWrapper.setProps({ className, id });
+        expect(mountedWrapper.hasClass(className)).toBeTruthy();
+        expect(mountedWrapper.find(`#${id}`).length).toBeGreaterThan(0);
     });
 
     it("Should render subheader is included in props", () => {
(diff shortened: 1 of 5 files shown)


src/Accordion/Accordion.tsx
```diff
@@ -71,7 +72,7 @@ const Accordion: React.FunctionComponent<AccordionProps> = (props: AccordionProp
     }
 
     return (
-        <div className={"custom-accordion" + (props.className ? ` ${props.className}` : "")} id={props.id} >
+        <div className={"custom-accordion" + (props.className ? ` ${props.className}` : "") + (props.alternative ? " alternate-accordion" : "")} id={props.id} >
```
DanSallau: Move this to a variable

```

let customClasses : string = "";
customClasses  = props.className ? props.className : "";

if(props.alterbative) {
    customClasses  += " " + props.alternative;
}

<div className={"custom-accordion" + customeClasses} id={props.id}>
``` 
Doing this would simplify things and ensure we don't keep elongating the className whenever we have another class to append.

src/Accordion/Accordion.tsx
```diff
@@ -27,35 +27,42 @@ export interface AccordionProps {
...
     const [active, setActive] = React.useState<number>(null);
-    const [className, setClassName] = React.useState<string>("custom-accordion");
+    const [itemClassName, setItemClassName] = React.useState<string>("custom-accordion");
     const [idList, setIdList] = React.useState<Array<string>>([]);
 
     React.useEffect(() => {
         constructIds();
-        constructClassName();
+        constructItemClassName();
     }, []);
```
DanSallau: @yousifalraheem why doesn't this have a dept too? Also how is it different from the useEffect below ?

src/Accordion/Accordion.tsx
```diff
@@ -85,7 +92,7 @@ const Accordion: React.FunctionComponent<AccordionProps> = (props: AccordionProp
                         role="button"
                     >
                         <div
-                            className={"header-wrapper" + (item.subHeaderText ? " with-sub-header" : "")}
+                            className={"header-wrapper" + (item.subHeaderText ? " with-sub-header" : "") + (props.alternative && !props.customIcon ? " alternate-header" : "")}
```
DanSallau: Id doable, move this to a variable too.

src/Accordion/Accordion.tsx
```diff
@@ -27,35 +27,42 @@ export interface AccordionProps {
...
         props.list.map(() => idListToSet.push(randomId("accordion-")));
         setIdList(idListToSet);
     }
 
-    function constructClassName(): void {
+    /**
+     * Constructs the `classname` to be used in accordion items
+     */
+    function constructItemClassName(): void {
         let cn: string = "accordion-item";
```
omarVodiaK: I suggest to add this as a utility, this function can help set className when the criterias are too much to track especially for handling spaces:
```
function classList(classes) {
  return Object
    .entries(classes)
    .filter(entry => entry[1])
    .map(entry => entry[0])
    .join(' ');
}

```
The logic is easy to understand:

1. First, it transforms our input object in an array of key-value pairs (one pair for each property).
2. Then, it filters the resulting array to eliminate items whose toggling criteria evaluate to false.
3. After that, it maps each item of the filtered array to its respective class name, outputting a new array with a list of classes.
4. Last, but not least, it joins all classes using a space as separator. The function returns the resulting string.

Running this will return => "accordion-item otherClass"
```
classNames({
  'accordion-item': true,
  'otherClass': true,
  'probablyNotClass': false,
})
```

src/Accordion/Accordion.tsx
yousifalraheem: An empty array means it will be executed when component mounts, I guess we don't need this since the next `useEffect` also gets called when component mounts. But still, this is @kherP's pull-request and I can solve it on my branch not this one. Thanks for noticing.

mario-subo: There is a big bug in here caused by not passing the dependencies to the dependency array of the React.useEffect. Since constructClassName() is being used inside the Effect and is reading from props like `props.iconPosition` and `props.iconLocation` they should be added to the dependencies. So basically it should look like this: 

```
    React.useEffect(() => {
        constructIds();
        constructClassName();
        constructItemClassName();
    }, [props.iconPosition, props.iconLocation]);
```

otherwise the effect will not be run again if iconPosition or location props change

You can read more in depth about it in this wonderful article by Dan Abramov:
https://overreacted.io/a-complete-guide-to-useeffect/#what-happens-when-dependencies-lie

mario-subo: Oh I see you added after I commented lol

src/Accordion/Accordion.tsx
kherP: new state is introduced to cater for multiple classes in component

src/Accordion/Accordion.tsx
kherP: removed the newly added condition as it is not necessary to set default icon for new theme

src/Accordion/Accordion.tsx
kherP: removed first useEffect as 2nd useEffect will call the same functions

src/Accordion/Accordion.tsx
kherP: new state is introduced to cater for this

src/Accordion/readme.md
```diff
@@ -38,6 +38,7 @@ These are the current available properties:
 | iconTransition?     | `string`                               | Icon transition rotation degree<sup>2</sup>                                             |
 | id?                 | `string`                               | Element id                                                                             |
 | list                | `Array<AccrodionListItem>`<sup>3</sup> | List of accordion items                                                                |
+| alternative?        | `boolean`                              | Toggle alternative style of accordion                                                  |
```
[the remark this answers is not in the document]
kherP: sorted in new commit

src/Accordion/Accordion.test.tsx
```diff
@@ -129,4 +129,14 @@ describe("Component: Accordion", () => {
         mountedWrapper.setProps({ list: [] });
         expect(mountedWrapper.find(".text-wrapper").length).toBeFalsy();
     });
+
+    describe("Should render alternate theme", () => {
```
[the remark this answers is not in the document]
kherP: replaced **alternate** with **alternative**

src/Accordion/Accordion.tsx
```diff
@@ -27,35 +27,57 @@ export interface AccordionProps {
...
         setIdList(idListToSet);
     }
 
+    /**
+     * Constructs the `classname` to be used in accordion wrapper
+     */
     function constructClassName(): void {
+        let cn: string = "custom-accordion";
+        cn  += props.className ? ` ${props.className}` : "";
+        if (props.alternative) {
```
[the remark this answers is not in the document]
kherP: removed if condition and replaced with ternary operator

PR #68, opened 2019-10-13
Enhancement/accordion with ref

DanSallau:
Instead of relying of maxHeight just for the the sake of transition, i think we should dynamically calculate them and allow the div takes it original or auto height where necessary like below. What do you think ?


src/Accordion/Accordion.tsx
```diff
@@ -30,40 +25,45 @@ export interface AccordionProps {
...
+interface TransitionProps {
+    height?: string;
+    transition?: string;
+    dataCollapsed?: boolean;
+    ariaExpanded?: boolean;
+}
+
 const Accordion: React.FunctionComponent<AccordionProps> = (props: AccordionProps) => {
+    const collapsableRef: React.MutableRefObject<Array<React.RefObject<HTMLDivElement>>> = React.useRef(props.list.map(() => React.createRef<HTMLDivElement>()));
+    const [transitionProperties, setTransitionProperty] = React.useState<Array<TransitionProps>>(() => props.list.map((): TransitionProps => ({
```
kherP: it's better to make `transitionProperties` state variable consistent. In this case, `setTransitionProperty` should be `setTransitionProperties`

DanSallau: @kherP setTransitionProperty  update the transition property of a single item in the array of items , hence the singularity in the wording.

PR #73, opened 2019-10-18
Feature/table

### LINKED ISSUE #39: Table component
### Table component description
- [x] Pagination
- [x] Filters
- [x] Search
- [x] Sort
- [x] Collapsable
- [ ] Direct data manipulation (editable fields)
- [ ] Different sizes (Normal, condensed, etc)
- [ ] Different themes (based on SEB's bootstrap)

### LINKED ISSUE #62: Table design (update)
### Component review 2019-10-08
There are two small and important changes on table design:
- Bigger table header font: 16 px, medium, Grey 2 (#333333)
- New sorting icon: Normal = Grey 7 (#B2B3B3), Selected= Dark Blue 2 (#007AC7), Hover = Blue (#41B0EE)
- The sorting icon is FontAwesome solid "chevron-up" and solid "chevron-down", on top of each other, width 10 x height 16 px (together)

image

There is more to come, but these changes are the most urgent. So the rest of what we know about tables stays the same for now.

Thank you!
/Ulrika
Design Management

DanSallau:
fixes #39
fixes #62 

### **Preview**

Final solution prebiew
Nice

### Requirement from design library team
tabell 2019-10-08


package.json
```diff
@@ -80,6 +80,7 @@
 		"react-router": "^5.1.1",
 		"react-router-dom": "^5.1.1",
 		"react-slick": "^0.25.2",
+		"react-table": "^7.0.0-beta.12",
```
yousifalraheem: You should not build your component on a `beta` version of a dependency. Try installing a `stable` version instead

DanSallau: We are removing this dependency . Building our own.

mario-subo:
There's a bug when you hover over the ">>" pagination button the tooltip says {prop.lastText}

mario-subo:
In sorting mode, clicking on the header text should also trigger sort. Not just pressing the sort icon.

mario-subo:
The elipsis menu should not open on hover but on click

mario-subo:
image

src/Table/readme.md
```diff
@@ -0,0 +1,83 @@
...
+    footer={
+        <Pagination
+            value={paginationValue}
+            onChange={setPagination}
+            size={3}
+            offset={3}
+            useFirstAndLast={true}
+        />
+    }
+/>
```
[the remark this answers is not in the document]
DanSallau: We have the basic use example to the top so this is in order.

DanSallau:
> mode

I intend to have a grouping too initially hence the reason why sorting on icon click instead of on column header clicked . The grouping comes in the next version and still column find a place to place the icon beside the header. If the two sorting and grouping are in the header as we intend to, it's counterproductive to have the sorting on header click, me thing.

DanSallau:
> The elipsis menu should not open on hover but on click

Hmm, Why ? Do you have any example or reason ?

DanSallau:
> There's a bug when you hover over the ">>" pagination button the tooltip says {prop.lastText}

To be fixed by https://github.com/sebgroup/react-components/pull/100

mario-subo:
> > The elipsis menu should not open on hover but on click
> 
> Hmm, Why ? Do you have any example or reason ?

Maybe you can ask Boon Ying as well? I just think it's common practice to open on click especially if it's elipsis. Would you as a user just looking at the button expect it to open on hover?

src/Table/Table.tsx
```diff
@@ -0,0 +1,809 @@
...
+                setAllRowsChecked(false);
+            } else {
+                setAllRowsChecked(true);
+            }
+        }
+    }, [currentTableRows]);
+
+    React.useEffect(() => {
+        if (!!props.searchProps?.onSearch) {
+            if (props.searchProps.triggerSearchOn === "Change") {
```
kherP: i think u can just use one if with && here

src/Table/Table.tsx
```diff
@@ -0,0 +1,809 @@
...
+                doSearch();
+            }
+        }
+    }, [props.searchProps?.searchInColumns, props.searchProps?.searchText]);
+
+    React.useEffect(() => {
+        if (!!props.searchProps?.onSearch) {
+            if (props.searchProps?.triggerSearchOn === "Submit") {
```
kherP: i think u can just use one if with && here

src/Table/Table.tsx
DanSallau: Yes, among the props i believe there is onSort callback that update the user with the newly sorted rows.

kherP: > Yes, among the props i believe there is onSort callback that update the user with the newly sorted rows.

that onSort i think instead of returning the sorted rows, u should just return an object with { type: "desc|asc", columnName: "col1" } something like this. The  reason is if the user is using the Pagination onChange callback to get data for page 2, and the data loaded, when u sort the column, it should sort the whole bunch of data instead of just the data of page 2

src/Table/Table.test.tsx
```diff
@@ -0,0 +1,314 @@
...
+                <Table
+                    columns={columns}
+                    data={smallData}
+                    actionLinks={actionLinks}
+                />, container
+            );
+        });
+
+        // trigger and open action column
+        expect(container.querySelector("tbody tr.parent-row > td .action-column .ellipsis-dropdown-holder .dropdown-content.active")).toBeNull();
```
mario-subo:
```suggestion
        const target: Element = container.querySelector("tbody tr.parent-row > td .action-column .ellipsis-dropdown-holder .dropdown-content.active");
        expect(target).toBeNull();
```

replace all duplicated code with target to keep it DRY

src/Table/Table.tsx
DanSallau: They are individual checking kp. ` `props.searchProps?.triggerSearchOn === "Submit"` is the same as props.searchProps && props.searchProps.triggerSearchOn === "Submit" `

develop/components/pages/TablePage.tsx
```diff
@@ -0,0 +1,228 @@
...
+                                    size={listSize}
+                                    useFirstAndLast={true}
+                                />
+                            }
+                        />
+                    </div>
+
+                    <p>Here is an example with expandable subrows and rowDetails</p>
+                    <div className="result wide">
+                        <Table
```
omarVodiaK: This table have a bug, those are the steps to reproduce:
1. expand the first item in the row.
2. expand any item if the subrow
3. shrink the row.
the details "Lorem ipsum" is still displayed.

Im not sure if this is the expected behavior.

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+/**
+ * sum the total of columns or cols in a row
+ * @param colsLength the length of the columns
+ * @param useSelection add a column for selection checkboxes
+ * @param useShowActionColumn add another column for action columns
+ * @param useGroupBy add another columns for groupby
+ */
+function sumCols(colsLength: number, useSelection?: boolean, useShowActionColumn?: boolean, useGroupBy?: boolean): number {
+    let sum = colsLength;
+    if (useSelection || useGroupBy) {
```
omarVodiaK: Can this be written like this
```suggestion
    if (useSelection || useGroupBy || useShowActionColumn) { 
    sum = sum+1;
    }
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+    selectedRow: TableRow;
+    tableRef: React.RefObject<HTMLTableElement>;
+}
+
+const ActionColumn: React.FunctionComponent<ActionColumnProps> = (props: ActionColumnProps) => {
+    const btnPrimaryRandomIds = randomId("btn");
+    const [dropup, setDropup] = React.useState<boolean>(false);
+    const actionRef: React.RefObject<HTMLDivElement> = React.createRef<HTMLDivElement>();
+
+    function getActionColumnClasses() {
```
omarVodiaK:
```suggestion
    function getActionColumnClasses(): string
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+        }
+
+        return className;
+    }
+
+    return (
+        <div className="action-column">
+            {props.primaryActionButton &&
+                <button
+                    id={btnPrimaryRandomIds}
```
omarVodiaK: This shouldn't random, should be possible to pass the id as a property, setting it as random may create tracking issues

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+                    type="button"
+                    className="btn btn-outline-primary btn-sm"
```
omarVodiaK: same for className

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+            className += " dropup";
+        }
+
+        return className;
+    }
+
+    return (
+        <div className="action-column">
+            {props.primaryActionButton &&
+                <button
```
omarVodiaK: is it possible to use button component ?

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+                    type="button"
+                    className="btn btn-outline-primary btn-sm"
+                    onClick={(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+                        props.primaryActionButton.onClick(e, props.selectedRow);
+                    }}
+                >
+                    {props.primaryActionButton.label}
+                </button>
+            }
+            {props.actionLinks && props.actionLinks.length > 0 &&
```
omarVodiaK:
```suggestion
            {props.actionLinks && props.actionLinks.length &&
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
+import * as React from "react";
+import { randomId } from "../__utils/randomId";
+
+import "./table-style.scss";
+
+const angleDown: JSX.Element = <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 512"><path d="M119.5 326.9L3.5 209.1c-4.7-4.7-4.7-12.3 0-17l7.1-7.1c4.7-4.7 12.3-4.7 17 0L128 287.3l100.4-102.2c4.7-4.7 12.3-4.7 17 0l7.1 7.1c4.7 4.7 4.7 12.3 0 17L136.5 327c-4.7 4.6-12.3 4.6-17-.1z" /></svg>;
+const angleRightIcon: JSX.Element = <svg name="angle-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 192 512"><path d="M166.9 264.5l-117.8 116c-4.7 4.7-12.3 4.7-17 0l-7.1-7.1c-4.7-4.7-4.7-12.3 0-17L127.3 256 25.1 155.6c-4.7-4.7-4.7-12.3 0-17l7.1-7.1c4.7-4.7 12.3-4.7 17 0l117.8 116c4.6 4.7 4.6 12.3-.1 17z" /></svg>;
+const ellipsis: JSX.Element = <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512"><path d="M192 256c0 17.7-14.3 32-32 32s-32-14.3-32-32 14.3-32 32-32 32 14.3 32 32zm88-32c-17.7 0-32 14.3-32 32s14.3 32 32 32 32-14.3 32-32-14.3-32-32-32zm-240 0c-17.7 0-32 14.3-32 32s14.3 32 32 32 32-14.3 32-32-14.3-32-32-32z" /></svg>;
```
omarVodiaK: you dont need ellipsis here, the icon was introduced with @sebgroup/bootstrap v5

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+const sortAsc: JSX.Element = <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 644"><path fill="#adadad" transform="translate(0 240)" d="M207.029 381.476L12.686 187.132c-9.373-9.373-9.373-24.569 0-33.941l22.667-22.667c9.357-9.357 24.522-9.375 33.901-.04L224 284.505l154.745-154.021c9.379-9.335 24.544-9.317 33.901.04l22.667 22.667c9.373 9.373 9.373 24.569 0 33.941L240.971 381.476c-9.373 9.372-24.569 9.372-33.942 0z" /><path transform="translate(0 -100)" d="M240.971 130.524l194.343 194.343c9.373 9.373 9.373 24.569 0 33.941l-22.667 22.667c-9.357 9.357-24.522 9.375-33.901.04L224 227.495 69.255 381.516c-9.379 9.335-24.544 9.317-33.901-.04l-22.667-22.667c-9.373-9.373-9.373-24.569 0-33.941L207.03 130.525c9.372-9.373 24.568-9.373 33.941-.001z" /></svg>;
```
omarVodiaK: you dont need sortAsc here, the icon was introduced with @sebgroup/bootstrap v5

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+const sortDesc: JSX.Element = <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 644"><path transform="translate(0 240)" d="M207.029 381.476L12.686 187.132c-9.373-9.373-9.373-24.569 0-33.941l22.667-22.667c9.357-9.357 24.522-9.375 33.901-.04L224 284.505l154.745-154.021c9.379-9.335 24.544-9.317 33.901.04l22.667 22.667c9.373 9.373 9.373 24.569 0 33.941L240.971 381.476c-9.373 9.372-24.569 9.372-33.942 0z" /><path fill="#adadad" transform="translate(0 -100)" d="M240.971 130.524l194.343 194.343c9.373 9.373 9.373 24.569 0 33.941l-22.667 22.667c-9.357 9.357-24.522 9.375-33.901.04L224 227.495 69.255 381.516c-9.379 9.335-24.544 9.317-33.901-.04l-22.667-22.667c-9.373-9.373-9.373-24.569 0-33.941L207.03 130.525c9.372-9.373 24.568-9.373 33.941-.001z" /></svg>;
```
omarVodiaK: you dont need sortDesc here, the icon was introduced with @sebgroup/bootstrap v5

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+    const tableRef: React.RefObject<HTMLTableElement> = React.createRef<HTMLTableElement>();
+    return (
+        <div className={"table-responsive" + (props.loading ? " skeleton-loader skeleton-loader-table" : "")}>
+            <table className={"table" + (props.className ? ` ${props.className}` : "")} ref={tableRef}>
+                <thead>
+                    <tr>
+                        {props.useRowSelection ?
+                            <th>
+                                <div className="custom-control custom-checkbox">
+                                    <input
```
omarVodiaK: Would be possible to use CheckBox component here ?

src/Table/Table.tsx
DanSallau: No it cannot be, They are independent and each has its bearing.

src/Table/Table.tsx
DanSallau: No it cannot be tracked and thus random fit it. The actions are replicated across rows and hence allowing user to pass the ids will result in duplications.

src/Table/Table.tsx
DanSallau: I disagree with this too. Considering each row renders this , i don't see a use for a className. If the user want to customise this , he can rely on his custom className and the last action column, which is one thing the buttons have in common. I don't foresee a situation where the user customise one button leaving the rest.

src/Table/Table.tsx
DanSallau: Nope, its a component and should be independent. One thing that can be utilise here is the bootstrap classes which we did above.

src/Table/Table.tsx
DanSallau: Nope, Your condition will be true for empty array

src/Table/Table.tsx
```diff
@@ -0,0 +1,977 @@
...
+                                setDropup(false);
+                            }
+                        } else {
+                            setDropup(false);
+                        }
+                        e.preventDefault();
+                        props.onActionDropped(e);
+                    }}
+                >
+                    <div className="icon-holder" id={"ellipsis-" + props.selectedRow.rowIndex}>
```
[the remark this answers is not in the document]
DanSallau: We need it omar because we can't count on all our users using seb-bootstrap.

src/Table/Table.tsx
DanSallau: We can't count on all our users to use seb-bootstrap hence the ellipsis here.

src/Table/Table.tsx
DanSallau: I think we should have ours as a component. we shouldn't rely on that and the fact that all our users will use seb-bootsrap.

src/Table/Table.tsx
DanSallau: Difference component @omarVodiaK . you can't mix the two.

src/Table/Table.tsx
DanSallau: I believe you are confusing backend pagination with front end one. You will not be needing The pagination here when you have an api that support the backend pagination.

src/Table/Table.tsx
omarVodiaK: for example if i call sumCols function with this parameters:
```
sumCols(colsLength = 3, useSelection = true, useShowActionColumn  = true, useGroupBy  = true)
sumCols(colsLength = 3, useSelection = false, useShowActionColumn  = true, useGroupBy  = true)
```
What is the expected return value for each case

src/Table/Table.tsx
omarVodiaK: My Point is you are using a random function to generate a random number and it makes it hard to predict which button is which id because it looks like this **id=btn1579229636563.0505**, the ActionColumn component is called inside props.rows.map so you can use the index parameter to define a unique id for each button.

This way tracking for the first or last element of the table would be possible.

Same goes for subrows.

src/Table/Table.tsx
omarVodiaK: [].length ? true : false  returns false

src/Table/Table.tsx
```diff
@@ -0,0 +1,984 @@
...
+                                                }}
+                                            />
+                                        </td>
+                                    }
+                                </tr>
+
+                                {row.subRows.map((subRow: TableRow) => {
+                                    const checkSubRowRandomIds: string = randomId("sub-chk-");
+                                    return (
+                                        <React.Fragment key={`sub-row-${subRow.rowIndex}`}>
```
omarVodiaK: The row and subrow parts looks like duplicated code, it will be better to create a tableRow component that can act as a row or a subrow.

src/Table/Table.tsx
```diff
@@ -0,0 +1,984 @@
...
+interface TableUIProps {
+    actionLinks?: Array<ActionLinkItem>;
+    allRowsAreSelected?: boolean;
+    className: string;
+    columns: Array<TableHeader>;
+    footer: React.ReactNode;
+    loading: boolean;
+    onActionDropped: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, row: TableRow, rowIndex?: number) => void;
+    onAllItemsSelected?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    onItemSelected?: (e: React.ChangeEvent<HTMLInputElement>, row: TableRow, type: "row" | "subRow", rowIndex?: number) => void;
```
omarVodiaK: type: "row" | "subRow" should be created as a type and used across the component

src/Table/Table.tsx
omarVodiaK: i think what she meant is 
```suggestion
            if (!!props.searchProps?.onSearch && props.searchProps?.triggerSearchOn === "Submit") { doSearch() }
```

src/Table/Table.tsx
DanSallau: Plus one for every true

src/Table/Table.tsx
DanSallau: According to your code above, initializing the array with empty [] will result in truthy.  I think relying on the length is better.

src/Table/Table.tsx
DanSallau: Thought about that, you might noticed i took the actionComponent out and make it into another FC. However, while they look familiar. They are kind of different in so many ways ranging from classNames to styling and paddings . While i agree with you that they have a lot in common.  Reusing a component for the two will introduce complications and will make it difficult to maintain. Maintenance-wise, the verbose version is clearer.

src/Table/Table.tsx
DanSallau: My point is tracking isn't a primary hear @omarVodiaK hence the use of the random. If we want to generate unique keys in a situation where the identity of the subject is not needed, i think the random method gives more uniqueness to the id.

src/Table/Table.tsx
mario-subo: image

src/Table/Table.tsx
omarVodiaK: you are right it gives a uniqueness but you cant be sure if it will be always unique when you use random, it might be a case even if its and edge case where randomId function will return the same result for two elements, especially if the table contains many rows.

Also the current id is useless because:

- no one can predict what is its value (getElementbyId is not possible)
-  it doesnt help for accessibility (no aria-*** attributes defined)

So my suggestion is either remove that part, or use a simple one so that at least developers can use getElementbyId

develop/components/pages/TablePage.tsx
```diff
@@ -0,0 +1,228 @@
+import * as React from "react";
+import { Table, Column, TableRow, PrimaryActionButton, ActionLinkItem, TableHeader, DataItem } from "../../../src/Table";
+import makeData from "../../utils/makeData";
```
yousifalraheem: Shouldn't `utils` here be `__utils`?

develop/components/pages/TablePage.tsx
```diff
@@ -0,0 +1,228 @@
...
+                        />
+                    </div>
+
+                    <p>Here is an example with row selection and subRows</p>
+                    <div className="result wide">
+                        <Table
+                            columns={columns}
+                            data={smallData}
+                            onRowSelected={(rows: Array<TableRow>) => { console.log("The selected rows are ", rows); }}
+                            onRowExpanded={(rows: Array<TableRow>) => { console.log("the expanded ros are ", rows); }}
```
yousifalraheem: I noticed you have a lot of console logs in this file. Console logs serve no purpose for the majority of the users. You shouldn't expect them to open the console to verify that certain aspects of the component are triggering the right methods. Scan this file and remove them all.

package.json
```diff
@@ -66,76 +66,78 @@
 		"package-lock-sanitizer": "package-lock-sanitizer"
 	},
 	"dependencies": {
-		"@sebgroup/bootstrap": "^4.0.0",
-		"chart.js": "^2.8.0",
+		"@sebgroup/bootstrap": "^5.1.0",
```
yousifalraheem: Do you need this? this is an upgrade to bootstrap which means we need to test that all the other component still working properly

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+ * @param items table rows array
+ * @param columnName the target column name
+ * @param sortDirection the sort direction
+ */
+function sortArray(items: Array<TableRow> = [], columnName: string, sortDirection: sortDirectionTypes): Array<TableRow> {
+    const sortedItems: Array<any> = [...items].sort((firstItem: TableRow, secondItem: TableRow) => {
+        let result: number = 0;
+        if (sortDirection === sortDirectionTypes.Ascending) {
+            if (isNaN(secondItem[columnName]) && isNaN(firstItem[columnName])) {
+                result = String(firstItem[columnName]).localeCompare(String(secondItem[columnName]), ["sw", "en"], { sensitivity: "base", ignorePunctuation: true });
```
yousifalraheem: Why is Swedish hardcoded here? this component is open source. What if the user is from German. Maybe should utilize `navigator.languages` or allow the user to pass the locale

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+interface ActionColumnProps {
+    actionLinks?: Array<ActionLinkItem>;
+    onActionDropped?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+    primaryActionButton?: PrimaryActionButton;
+    selectedRow: TableRow;
+    tableRef: React.RefObject<HTMLTableElement>;
+}
+
+const ActionColumn: React.FunctionComponent<ActionColumnProps> = (props: ActionColumnProps) => {
+    const btnPrimaryRandomIds = randomId("btn");
```
yousifalraheem:
```suggestion
    const btnPrimaryRandomIds: string = randomId("btn");
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                        props.primaryActionButton.onClick(e, props.selectedRow);
+                    }}
+                >
+                    {props.primaryActionButton.label}
+                </button>
+            }
+            {props.actionLinks && props.actionLinks.length > 0 &&
+                <div
+                    className="ellipsis-dropdown-holder"
+                    onClick={(e) => {
```
yousifalraheem:
```suggestion
                    onClick={(e: React.MouseEvent<HTMLDivElement>) => {
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+
+    return (
+        <div className="action-column">
+            {props.primaryActionButton &&
+                <button
+                    id={btnPrimaryRandomIds}
+                    type="button"
+                    className="btn btn-outline-primary btn-sm"
+                    onClick={(e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+                        props.primaryActionButton.onClick(e, props.selectedRow);
```
yousifalraheem: since this is coming from the props you might wanna verify it has a value before executing
```suggestion
                        props.primaryActionButton?.onClick && props.primaryActionButton.onClick(e, props.selectedRow);
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                        }
+                        e.preventDefault();
+                        props.onActionDropped(e);
+                    }}
+                >
+                    <div className="icon-holder" id={"ellipsis-" + props.selectedRow.rowIndex}>
+                        {ellipsis}
+                    </div>
+                    <div
+                        className={getActionColumnClasses()}
```
yousifalraheem: create a state with an effect for `props.selectedRow` instead of recalculating with every interaction

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                        ref={actionRef}
+                    >
+                        {props.actionLinks.map((link: ActionLinkItem, index: number) =>
+                            <a
+                                href="#"
+                                key={`${link.label.replace(" ", "-")} - ${index}}`}
```
yousifalraheem: is this calculation really necessary? the key is only used by react, why can't you just use the index

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                    <div className="icon-holder" id={"ellipsis-" + props.selectedRow.rowIndex}>
+                        {ellipsis}
+                    </div>
+                    <div
+                        className={getActionColumnClasses()}
+                        ref={actionRef}
+                    >
+                        {props.actionLinks.map((link: ActionLinkItem, index: number) =>
+                            <a
+                                href="#"
```
yousifalraheem: you should not have a `#` as a URL since it might cause a reload or unwanted scroll jumps you can just remove it and add the styles necessary

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                                <tr className={"parent-row" + (row.expanded ? " expanded" : "")}>
+                                    {props.useRowSelection ?
+                                        <td className="row-selections-column">
+                                            <div className="custom-control custom-checkbox">
+                                                <input
+                                                    type="checkbox"
+                                                    className="custom-control-input"
+                                                    id={checkRowRandomIds}
+                                                    checked={row.selected}
+                                                    onChange={(e: React.ChangeEvent<HTMLInputElement>) => { props.onItemSelected(e, row, "row"); }}
```
yousifalraheem:
```suggestion
                                                    onChange={(e: React.ChangeEvent<HTMLInputElement>) => { props.onItemSelected && props.onItemSelected(e, row, "row"); }}
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                                                    {row.expanded ? angleDown : angleRightIcon}
+                                                </div>
+                                            }
+                                        </td>
+                                        :
+                                        ((row.subRows.length > 0 || row.rowContentDetail) && props.rowsAreCollapsable) &&
+                                        <td>
+                                            <div
+                                                className={"icon-holder" + (row.expanded ? " active" : "")}
+                                                onClick={(e: React.MouseEvent<HTMLDivElement, MouseEvent>) => { props.onRowExpanded(e, row); }}
```
yousifalraheem:
```suggestion
                                                onClick={(e: React.MouseEvent<HTMLDivElement, MouseEvent>) => { props.onRowExpanded && props.onRowExpanded(e, row); }}
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                                    })}
+                                    {props.useShowActionColumn &&
+                                        <td>
+                                            <ActionColumn
+                                                actionLinks={props.actionLinks}
+                                                primaryActionButton={props.primaryActionButton}
+                                                selectedRow={row}
+                                                tableRef={tableRef}
+                                                onActionDropped={(event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+                                                    props.onActionDropped(event, row, null);
```
yousifalraheem:
```suggestion
                                                    props.onActionDropped && props.onActionDropped(event, row, null);
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                                    const checkSubRowRandomIds: string = randomId("sub-chk-");
+                                    return (
+                                        <React.Fragment key={`sub-row-${subRow.rowIndex}`}>
+                                            <tr
+                                                className={"sub-row" + (subRow.expanded ? " expanded" : "")}
+                                                style={{ display: row.expanded ? "table-row" : "none" }}
+                                            >
+                                                {props.useRowSelection ?
+                                                    <td className="row-selections-column">
+                                                        <div className="custom-control custom-checkbox" style={{ marginLeft: "20px" }}>
```
yousifalraheem: Why is this hardcoded in JS? `style={{ marginLeft: "20px" }}` Just add it to the stylesheet. This will make it difficult for users to re=style the component. The only solution for them might be using `!important`

I found a bunch of these around this file. Just find them and remove them from JS and add them to CSS

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+                                                    );
+                                                })}
+                                                {props.useShowActionColumn && <td>
+                                                    <ActionColumn
+                                                        actionLinks={props.actionLinks}
+                                                        primaryActionButton={props.primaryActionButton}
+                                                        selectedRow={subRow}
+                                                        tableRef={tableRef}
+                                                        onActionDropped={(event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+                                                            props.onActionDropped(event, subRow, row.rowIndex);
```
yousifalraheem:
```suggestion
                                                            props.onActionDropped && props.onActionDropped(event, subRow, row.rowIndex);
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+
+    /**
+     * Call when item is selected
+     * @param e change event
+     * @param selectedRow The selected row
+     * @param type The row type (i.e either a row or subRow)
+     * @param rowIndex The index of the parent row incase of subRow
+     */
+    const onItemSelected = (e: React.ChangeEvent<HTMLInputElement>, selectedRow: TableRow, type: "subRow" | "row", rowIndex?: number): void => {
+        const updatedOriginalRows = selectItems(e.target.checked, tableRows, selectedRow, rowIndex, type);
```
yousifalraheem: missing type

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+        props.onRowSelected(selectedRowList);
+    };
+
+    /**
+     *
+     * @param exchange
+     * Called onAllItemsSelected
+     */
+    const onAllItemsSelected = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        const updatedOriginalRows = tableRows.map((originalRow: TableRow) => {
```
yousifalraheem: missing type

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+    };
+
+    /**
+     * Called on sub row is clicked
+     * @param e change ve
+     * @param row The subrow Selected row
+     * @param rowIndex The parent row index
+     */
+    const onSubRowExpanded = (e: React.MouseEvent<HTMLDivElement, MouseEvent>, row: TableRow, rowIndex: number): void => {
+        const updatedOriginalRows = tableRows.map((originalRow: TableRow) => {
```
yousifalraheem: missing type

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+        props.onRowExpanded(expandedRowList);
+    };
+
+    /**
+     *
+     * @param e change event
+     * @param row The selected row
+     */
+    const onRowExpanded = (e: React.MouseEvent<HTMLDivElement, MouseEvent>, row: TableRow): void => {
+        const updatedOriginalRows = tableRows.map((originalRow: TableRow) => {
```
yousifalraheem: missing type

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+
+    // useEffects ----------------------------------------------------------
+
+    // Adding event listener to listen to clicks outside the component on mount, removing on unmount
+    React.useEffect(() => {
+        document.addEventListener("mousedown", onClickOutside);
+        return () => {
+            document.removeEventListener("mousedown", onClickOutside);
+        };
+    });
```
yousifalraheem: You need to pass an empty array to `useEffect` if you only need it to happen when it **mount** and **unmount**
```suggestion
    }, []);
```

src/Table/Table.tsx
```diff
@@ -0,0 +1,985 @@
...
+
+    React.useEffect(() => {
+        doPaginate();
+    }, [tableRows]);
+
+    React.useEffect(() => {
+        if (!!props.onRowSelected) {
+            const notAllsAreRowsSelected = tableRows.some((row: TableRow) => !row.selected);
```
yousifalraheem: missing type

src/Table/index.ts
```diff
@@ -0,0 +1,6 @@
+
+(window as any).global = window;
+// @ts-ignore
+window.Buffer = window.Buffer || require("buffer").Buffer;
```
yousifalraheem: why? can you put a comment describing why we need this

src/Table/Table.tsx
DanSallau: I don't recommend exposing the locale as an input or params for users to pass. I feel like it will raise a lot of questions on what does it do and what is it for. It appearance in this domain will send a wrong signal to the user, thinking you can handle some kind of translation or language service for him. However, i do agree on using language locale and having a default where not available.

src/Table/index.ts
DanSallau: Oh, the recent node release has removed the array buffer polyfill i believe. It has been moved to a separate dependency  [here](https://www.npmjs.com/package/buffer) and you need to install it if you need it. So the code here simply check if the browser has it, otherwise require it.

package.json
mario-subo: I suggest you wait for #115  to be merged into develop and just pull the latest from develop into this branch

src/Table/Table.tsx
DanSallau: We will do this refactoring in another PR @omarVodiaK .

src/Table/Table.tsx
DanSallau: You are right, its not an id

omarVodiaK:
I believe we should merge after the change requests are done, we can publish it and continue with the improvements later.

PR #80, opened 2019-10-22
Bug/slider negative values

### LINKED ISSUE #78: Negative slider min acts very weird
MinusMin

**Describe the bug**
Negative value as min value makes the calculations realy strange on where the thumb will be located.

**To Reproduce**
1. Include react-components 2.0.3.
2. slider with min="-2", or "-3" or "-4" and max="7".
3. Slide the tumb back and forth.

**Expected behavior**
Slider thumb should follow mouse/finger.
Slider no longer than 100% width. Now it can end on for example 180% width.

**Desktop (please complete the following information):**
 - OS: Win 10
 - Browser: Chrome 77, Firefox 69, IE11

**Additional context**
Worked in react-components v1.1.0.

### LINKED ISSUE #79: Make slider accessible
Missing hover and focus state for slider component. Makes it hard if not impossible to follow the new accessibility law. People with disabilities have hard time to step through components when they do not look different when selected.

* Have highlights around thumb when it is selected (focused with tab/finger/mouse). Use box-shadow 3px lightblue like many other components (0.15s ease-in-out if fit). Probably use "focus" class for this.
* Make the thumb increase in size when hovering it (with a short animation). Extra size should be 5%. To compensate with the increased size make the non hovered state 5% smaller. See avanza.se for example (scroll down a bit to the green sliders).

Use number 1 of these:
SliderHighlights
SliderHover

**Additional context**
Ulrika Dettner-Larsson and Henrik Hoglund have approved this change.

### LINKED ISSUE #85: Slider startpoint
Be able to set where the blue bar should start from.

A setting that tells where the midpoint is. If not set it default to min value.
Especially useful when using min values less than 0 and want to show the blue bar from 0. 

Still use default behavior from before.

**Additional context**
SliderMid

Approved by: Ulrika Dettner-Larsson & Henrik Hoglund.

yousifalraheem:
Slider bug fixes:

- [x] Negative values. fixes #78
- [x] Focus and hover styles. fixes #79 
- [x] Disabled property in readme is labelled as string instead of boolean
- [x] Slider now extends to the sides
- [x] Active slider track fixed on zero as a startpoint. fixes #85 

ezgif-2-8208a5540365


src/Slider/Slider.tsx
```diff
@@ -50,29 +56,50 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
...
+     * @returns {number} The size
+     */
+    function getSize(minValue: number, maxValue: number): number {
+        if (maxValue > minValue) {
+            return maxValue - minValue;
+        } else { // Will calculate the size anyway, but it will show a warning since the min is larger than the max
+            console.warn(`The max value of the slider should be larger than the min value (Max:${max}, Min: ${min}`);
+            return minValue - maxValue;
+        }
+    }
```
mario-subo:
```
        if (maxValue <= minValue) {
            console.warn(`The max value of the slider should be larger than the min value (Max:${max}, Min: ${min}`);
        } 
        return minValue - maxValue;
```

src/Slider/Slider.tsx
```diff
@@ -50,29 +56,50 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
...
             return 0;
-        } else if (value > max) {
+        } else if (props.value >= max) {
             return 100;
         } else {
-            const offset: number = Math.abs(0 - min);
-            const result: number = Math.abs(((value - min) / (max - offset)) * 100);
-            return result;
+            const distanceFromMin: number = Math.abs(props.value - min);
+            // Calculating the percentage using this equation (distanceFromMin/size) * 100
```
mario-subo: This comment is not necessary

src/Slider/Slider.tsx
```diff
@@ -101,19 +128,10 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
...
-                        />
-                        <div
-                            className="custom-slider-slider-after"
-                            style={{ width: (100 - getPercentage(props.value)) + "%" }}
-                        />
-                        <div
-                            className="custom-slider-thumb"
-                            style={{ left: getPercentage(props.value) + "%" }}
-                        >
+                    <div className={"custom-slider-track" + (size / props.step <= 30 ? " with-transitions" : "")}>
```
mario-subo: All this logic seems to be very custom and advanced to be hardcoded inline. Either extract it to a function with a descriptive name or try to make it more readable somehow.

src/Slider/Slider.tsx
yousifalraheem: Your suggested changes won't work since I'm returning a value in both `if` and `else` where each one is different

src/Slider/Slider.tsx
yousifalraheem: It's just to explain what's included in the calculation. Furthermore, I don't think this is worthy of a commit. Let's just leave it there.

src/Slider/Slider.tsx
yousifalraheem: moved the logic to a method

develop/components/pages/SliderPage.tsx
```diff
@@ -32,9 +32,14 @@ const SliderExamples: React.FunctionComponent = () => {
     const [disabledSlider1, setDisabledSlider1] = React.useState<number>(2500);
     const [sebSlider, setSebSlider] = React.useState<number>(0);
     const [sebSliderError, setSebSliderError] = React.useState<string>("");
+    const [diverse, setDiverse] = React.useState<number>(diverseLabels[0].position);
```
omarVodiaK: check if diverseLabels[0] exist before checking the position propety, otherwise it will throw "cannot find position of undefined "

src/Slider/Slider.tsx
```diff
@@ -35,11 +35,17 @@ export interface SliderProps {
...
     const [max, setMax] = React.useState<number>(props.max || 100);
+    const [size, setSize] = React.useState<number>(0);
     const [labelsPositions, setLabelsPositions] = React.useState<Array<string>>([]);
+    const [thumbPosition, setThumbPosition] = React.useState<number>(0);
 
     React.useEffect(() => {
-        setMin(props.min || 0);
-        setMax(props.max || 100);
+        // Checking if the min or max are not numbers, null value or undefined
+        const minValue: number = typeof props.min !== "number" ? 0 : props.min;
```
omarVodiaK: typeof NaN is also a number, maybe we should add IsNaN
```suggestion
        const minValue: number = (typeof props.min !== "number" && isNaN(props.min)) ? 0 : props.min;
```

src/Slider/Slider.tsx
```diff
@@ -50,34 +56,65 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
...
+     * Finds the size between two numbers
+     * @param {number} minValue The minimum value
+     * @param {number} maxValue The maximum value
+     * @returns {number} The size
+     */
+    function getSize(minValue: number, maxValue: number): number {
+        if (maxValue > minValue) {
+            return maxValue - minValue;
+        } else { // Will calculate the size anyway, but it will show a warning since the min is larger than the max
+            console.warn(`The max value of the slider should be larger than the min value (Max:${max}, Min: ${min}`);
```
omarVodiaK: This warning is not enough since it appears only on the console, we should move the warning to the error prop or make sure that the component doesnt accept min > max

develop/components/pages/SliderPage.tsx
yousifalraheem: This is the `sliderPage` file, and `diverseLabels` is a predefined constant array. It will always exist because it's part of the documentation

src/Slider/Slider.tsx
yousifalraheem: `typeof variable` is a more accurate than `isNaN` to what I'm trying to achieve given the following:

| value | `isNaN` | `typeof test !== "number"` |
| -- | -- | -- |
| `null` | `false` | `true` |
| `"1"` | `false` | `true` |
| `"+1"` | `false` | `true` |
| `""` | `false` | `true` |
| `new Date()` | `false` | `true` |
| `[]` | `false` | `true` |

src/Slider/Slider.tsx
yousifalraheem: The error prop is controlled by the parent component so we can't use that. I only put a warning because I thought this should be something that is caught during development. The calculation will proceed with no issue but it shouldn't happen. The only scenario I can think of is when someone mistakenly thinks that the slider could go backward if it's in negative like setting the `min` to `0` and the `max` to `-30` thinking that it's a `from` - `to`. In this case, this warning will be shown.

PR #84, opened 2019-10-29
Bug/pagination

### LINKED ISSUE #81: Pagination not working with pageSize less than offset default of 10
**Describe the bug**
The pagination seem to be omitting pageSize and generating negetive values. 

**To Reproduce**
```tsx
<Pagination
    value={paginationValue}
    onChange={setPagination}
    size={3}
    useFirstAndLast={true}
/>
```

**Screenshots**
image

**Additional context**
The `getlist ` should be calculated once and only when there is a change in value.

DanSallau:
fixes #81


src/Pagination/Pagination.tsx
```diff
@@ -33,27 +35,37 @@ export const Pagination: React.FunctionComponent<PaginationProps> = React.memo((
      * @param {number} offset The offset to show the numbers
```
mario-subo: JSdocs need to reflect the changes

src/Pagination/Pagination.tsx
```diff
@@ -33,27 +35,37 @@ export const Pagination: React.FunctionComponent<PaginationProps> = React.memo((
...
-                list.push(i);
-            }
-        } else {
-            range.min = value + 1 - (Math.ceil(offset / 2));
-            range.max = range.min + offset - 1;
-            for (let i = range.min; i <= range.max; i++) {
-                list.push(i);
+
+        // generate the array
+        for (let i = 1; i <= pagingSize; i++) {
```
mario-subo: For consistency consider deciding if you want to either send all outside values into this function as a param or just use the values directly.  Right now you're sending the `value` and `length` as params but `pagingSize` is being used directly. 

You can also consider building the list and setting it to an internal state within a React.useEffect which would run every time any of the values you need change. Then you would just render the list within the return. Either way you should clean it up a bit.

mario-subo:
I think there's also a bug with the next button where the icon doesn't turn white on hover, can you check it out?

src/Pagination/Pagination.tsx
```diff
@@ -81,7 +104,7 @@ export const Pagination: React.FunctionComponent<PaginationProps> = React.memo((
                                 </button>
                             </li>
                         }
-                        {getList(props.value, props.size, Initialoffset).map((num) => {
+                        {list.map((num) => {
```
yousifalraheem:
```suggestion
                        {list.map((num: number) => {
```

src/Pagination/Pagination.tsx
```diff
@@ -126,7 +149,7 @@ export const Pagination: React.FunctionComponent<PaginationProps> = React.memo((
                 }
                 {props.useDotNav &&
                     <ul className={"pagination dotnav"}>
-                        {getList(props.value, props.size, props.size).map((num) => {
+                        {list.map((num) => {
```
yousifalraheem:
```suggestion
                        {list.map((num: number) => {
```

src/Pagination/readme.md
```diff
@@ -44,6 +44,5 @@ These are the current available properties:
 | useFirstAndLast? | `boolean`                 | Use first and last navigation buttons (default: `false`) |
 | useTextNav?      | `boolean`                 | Use text-base navigation buttons (default: `false`)      |
 | value            | `number`                  | Current page                                             |
+| pagingLength?    | `number`                  | The length of the pagination or number of pages          |
 
-## Footnote
-1. Dot navigation does not support `offset`, it is not meant to have big size navigation. For that, use the normal numbered navigation.
```
yousifalraheem: I think you need to put back this footnote since it's important

src/Pagination/Pagination.tsx
```diff
@@ -23,37 +24,60 @@ export interface PaginationProps {
...
-            range.max = size;
-            for (let i = range.min; i <= range.max; i++) {
-                list.push(i);
-            }
-        } else {
-            range.min = value + 1 - (Math.ceil(offset / 2));
-            range.max = range.min + offset - 1;
-            for (let i = range.min; i <= range.max; i++) {
-                list.push(i);
+    function getList(value: number, length: number): void {
```
yousifalraheem: Don't you think a method called `getList` should return something? if not, then it shouldn't be called `getList` just to not confuse the reader

src/Pagination/Pagination.tsx
```diff
@@ -33,27 +35,37 @@ export const Pagination: React.FunctionComponent<PaginationProps> = React.memo((
...
-            for (let i = range.min; i <= range.max; i++) {
-                list.push(i);
+
+        // generate the array
+        for (let i = 1; i <= pagingSize; i++) {
+            list.push(i);
+        }
+
+        // the median value is what you use to obtain the centre of the page or equilibrium
+        const medianValue = Math.ceil(length / 2);
```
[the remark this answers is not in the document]
yousifalraheem: Why is this maked as resolved? The type still haven't been added

yousifalraheem:
Can you add more unit test? **Coveralls** is noticing a coverage decrease, this should be easily fixed

src/Pagination/Pagination.tsx
```diff
@@ -23,37 +24,59 @@ export interface PaginationProps {
...
-    const Initialoffset: number = props.offset ? props.offset : 10;
+
+    const [list, setList] = React.useState<Array<number>>([]);
+    const [dotnavList, setDotnavList] = React.useState<Array<number>>([]);
+    const [pagingSize, setPagingSize] = React.useState<number>(0);
+
+    React.useEffect(() => {
+        const initialOffset: number = props.offset ? props.offset : 10;
+        setPagingSize(Math.ceil(props.size / initialOffset));
+    }, [props.offset, props.pagingLength, props.size]);
```
omarVodiaK: is props.pagingLength required here, the effect would run if pagingLength changes.

src/Pagination/Pagination.tsx
```diff
@@ -23,37 +24,59 @@ export interface PaginationProps {
...
+
+    React.useEffect(() => {
+        const intialLength: number = props.pagingLength ? props.pagingLength : 5;
+        generateList(props.value, intialLength);
+    }, [pagingSize, props.value]);
```
omarVodiaK: maybe props.pagingLength should be added here.

omarVodiaK:
You should update the branch by merging latest develop

src/Pagination/Pagination.tsx
yousifalraheem:
```suggestion
    }, [props.offset, props.size]);
```

src/Pagination/Pagination.tsx
yousifalraheem:
```suggestion
    }, [pagingSize, props.value, props.pagingLength]);
```

src/Pagination/Pagination.tsx
DanSallau: nice catch, thanks.

PR #91, opened 2019-11-14
feat(dialogue): improve dialogue

yousifalraheem:
Adds the following:
- [X] Added disable states to dialogue buttons
- [X] Adds the ability to pass JSX elements or react components to the header or description props to be
rendered inside the dialogue


src/Dialogue/Dialogue.tsx
```diff
@@ -5,27 +5,28 @@ const TimesIcon: JSX.Element = <svg name="times" xmlns="http://www.w3.org/2000/s
 
 export interface DialogueProps {
     className?: string;
-    desc?: string;
+    desc?: string | JSX.Element | React.ReactNode;
+    disablePrimaryBtn?: boolean;
```
mario-subo: `disablePrimaryBtn` sounds like an action. You should name is `primaryBtnDisabled` instead.

Same for `disableSecondaryBtn`

PR #99, opened 2019-12-16
Refactor/tooltip

kherP:
### 1. introducing new attributes:
  - **trigger**:  tooltip trigger mode which takes ``click | hover | focus``
  - **content**: tooltip content which will replace **message**, **messageGroup**, **title** where users can pass their own custom element as tooltip content.
  - **onVisibleChange**: callback when tooltip visibility status change which will later replace **onClick**
  - **disableAutoPosition**: disable auto position for the tooltip and force tooltip to be at certain position.
  - **children**: tooltip reference where users are allowed to define their own tooltip reference other than icons, which will later replace **customSvg**

### 2. introducing auto position for the tooltip:
  - when the tooltip is overflowing the viewport, the tooltip will be placed in a new position automatically.

### 3. deprecating some existing attributes:
  - **customSvg**: will be replaced by **children**. Instead of just allowing icon as the reference of the tooltip, users can now set any element as the tooltip reference by using the **children** attribute
  - **message**: will be replaced by **content**. Users can choose to pass string or self-defined element as the tooltip content by using the **content** attribute.
  - **messageGroup**: will be replaced by **content**. Users can choose to pass string or self-defined element as the tooltip content by using the **content** attribute.
  - **title**: will be replaced by **content**. Users can choose to pass string or self-defined element as the tooltip content by using the **content** attribute.
  - **triggerOnHover**: will be replaced by **trigger**. Users can set the trigger mode of tooltip by using the **trigger** attribute.
  - **onClick**: will be replaced by **onVisibleChange**. As we are now introducing trigger as the attribute to set the trigger method of tooltip, the naming of onClick as the callback function of the tooltip visibility status change does not make sense anymore.
  - **width**: will be removed. Users do not have to set width for the tooltip.


develop/components/pages/TooltipPage.tsx
```diff
@@ -1,22 +1,81 @@
...
+            title: "Tooltip title",
+            message: "tooltip 1"
+        },
+        {
+            message: "tooltip 2"
+        }
+    ];
+    const defaultTooltipContent: string = "Tooltip content could be long, therefore, controlling the position and width is important";
+    const nodeTooltipContent: React.ReactNode = (<div><h1>Tooltip Header</h1><div>tooltip content</div></div>);
+    const [theme, setTheme] = React.useState<TooltipTheme>("primary");
```
omarVodiaK: would it be possible to use the value that you created in themeList, same thing for position and trigger

develop/components/pages/TooltipPage.tsx
```diff
@@ -1,22 +1,81 @@
...
+    const [customSvg, setCustomSvg] = React.useState<boolean>(false);
+    const [content, setContent] = React.useState<number>(0);
     let myTooltip: Tooltip;
 
     function dismissTooltip(e?: React.MouseEvent<HTMLDivElement>): void {
-        const dismissableTooltip: HTMLElement = document.getElementById("dismissable-tooltip");
-        if (event.target !== dismissableTooltip.firstChild) {
-            myTooltip.forceDismiss(e);
-        }
+        console.log("force dismissed");
```
omarVodiaK: remove logs

develop/components/pages/TooltipPage.tsx
```diff
@@ -30,33 +89,108 @@ const TooltipPage: React.FunctionComponent = () => {
...
-
-                    <p>Custom svg and can be dismissed when clicked outside</p>
-                    <div className="result">
-                        <Tooltip
-                            message="Tooltip message could be long, therefore, controlling the position and width is important"
-                            position="right"
-                            width={200}
-                            customSvg={moneySVG}
-                            id="dismissable-tooltip"
+                            content={content === 0 && (hasNodeAsContent ? nodeTooltipContent : defaultTooltipContent)}
```
omarVodiaK: its a bit confusing to have content as a number with values 1,2,3 its not clear what they mean, maybe you want to create an enum instead,

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+            tooltipRootRef = document.createElement("div");
+            tooltipRootRef.className = this.tooltipRootClassName;
+            document.body.appendChild(tooltipRootRef);
+        }
+        const newTooltip: HTMLDivElement = document.createElement("div");
+        newTooltip.className = `tooltip-content ${this.props.theme || "default"} ${this.props.position}`;
+        newTooltip.id = tooltipId;
+        newTooltip.setAttribute("role", "tooltip");
+        this.setTooltipContent(newTooltip);
+        tooltipRootRef.appendChild(newTooltip);
```
yousifalraheem: Nope

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
-                                <div key={index} className="message-list-item">
-                                    {item.title && <div className="title">{item.title}</div>}
-                                    {item.message && <div className="message">{item.message}</div>}
-                                </div>
-                            )}
-                        </div>
-                    }
-
-                    {(this.isPositioned("top") || this.isPositioned("right") || this.isPositioned("left")) && <div className="triangle" />}
+                    {this.props.children || <div className="icon">{this.props.customSvg ? this.props.customSvg : InfoCircleIcon}</div>}
```
yousifalraheem: We don't want to assume it's an icon. We want to start using the tooltip with any UI element. It could be an input element or anything else

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+                id={this.props.id}
+                ref={this.containerRef}
+            >
                 <div
-                    className="icon tooltip-icon"
-                    onClick={(e: React.MouseEvent<HTMLDivElement>) => !this.props.triggerOnHover && this.toggleTooltip(undefined, e)}
-                    onMouseEnter={() => this.props.triggerOnHover && this.toggleTooltip(true)}
-                    onMouseLeave={() => this.props.triggerOnHover && this.toggleTooltip(false)}
+                    className="tooltip-reference"
+                    aria-describedby={this.state.id}
```
yousifalraheem: Described by what? this should refer to a text element describing it. [See this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute)

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+                    tabIndex={-1}
+                    onClick={(e: React.MouseEvent<HTMLDivElement>) => ((!this.props.trigger && !this.props.triggerOnHover) || this.props.trigger === "click") && this.onTooltipToggle(e)}
+                    onMouseEnter={(e: React.MouseEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, true)}
+                    onMouseLeave={(e: React.MouseEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, false)}
+                    onFocus={(e: React.FocusEvent<HTMLDivElement>) => this.props.trigger === "focus" && this.onTooltipToggle(e)}
+                    onBlur={(e: React.FocusEvent<HTMLDivElement>) => this.onTooltipToggle(e, false)}
```
yousifalraheem: Move all the logic you have for this method `onTooltipToggle` to the method itself. You can know whether you should toggle `true` or `false` based on the `event.type`. Also when you are toggling it off in the `onBlur` method, you should check if it is already toggled off because setting the state will have a lot of effects which you don't want to have. Same goes for `onMouseLeave`

By the way, how is the `focus` and `blur` being triggered here? This is a `div` element with `tabIndex` set to `-1`?

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +12,92 @@ export interface TooltipMessageGroupItem {
...
+            tooltipContainer: null,
+            tooltipPositionChecker: null
         };
 
         this.forceDismiss = this.forceDismiss.bind(this);
-        this.toggleTooltip = this.toggleTooltip.bind(this);
+    }
+
+    componentDidMount() {
+        const randID: string = this.constructId();
```
yousifalraheem: move the logic of constructing the ID here since you never reuse it. And rename `id` in the state to `accessibilityId` or `containerId` or anything else that sets it apart from `this.props.id`

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +12,92 @@ export interface TooltipMessageGroupItem {
...
     title?: string;
+    /** @deprecated use trigger instead */
     triggerOnHover?: boolean;
+    /** @deprecated */
     width?: number;
+    content?: string | React.ReactNode;
+    trigger?: TooltipTrigger;
+    disableAutoPosition?: boolean;
+    onVisibleChange?: (event: React.MouseEvent<HTMLDivElement> | React.FocusEvent<HTMLElement>, visible: boolean) => void;
+    children?: React.ReactNode;
```
DanSallau: children here is a react reserve word and would be confusing to users. can you use a name that tallies with the variable intent ?

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+        }, () => {
+            this.state.tooltipPositionChecker.addTooltipContainer(this.state.tooltipContainer);
+        });
+    }
+
+    /**
+     * construct unique tooltip id
+     */
+    private constructId = (): string => {
+        const randId: string = randomId("tooltip-");
```
DanSallau: i believe `ramdomId` here comes from another dependency right ? unless its mention as a peer-dependency in this project, it may raise an issue;

develop/components/pages/TooltipPage.tsx
kherP: I was thinking to retain the console log as part of the example

develop/components/pages/TooltipPage.tsx
```diff
@@ -1,22 +1,81 @@
...
-        const dismissableTooltip: HTMLElement = document.getElementById("dismissable-tooltip");
-        if (event.target !== dismissableTooltip.firstChild) {
-            myTooltip.forceDismiss(e);
-        }
+        console.log("force dismissed");
+        myTooltip.forceDismiss(e);
+    }
+
+    function forceShowTooltip(e?: React.MouseEvent<HTMLDivElement>): void {
+        console.log("force showed");
```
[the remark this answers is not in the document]
kherP: I was thinking to retain the console log as part of the example

develop/components/pages/TooltipPage.tsx
kherP: added enum

src/Tooltip/Tooltip.tsx
kherP: described by the tooltip

src/Tooltip/Tooltip.tsx
kherP: by setting tabIndex to -1, it will allow the div to be focusable but not affecting the sequential keyboard navigation, you can refer more on this [link](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

src/Tooltip/Tooltip.tsx
kherP: removing the children attribute as the children here is the actual children of the component. I added the attribute to the props interface as it shows error when the component is written as function component

src/Tooltip/Tooltip.tsx
kherP: randomId is in __utils folder

src/Tooltip/Tooltip.tsx
kherP: renaming state.id to state.accessibilityId

src/Tooltip/Tooltip.tsx
kherP: Ya, we do not assume it as an icon. Users are now encouraged to define the tooltip reference as the children of the component

develop/components/pages/TooltipPage.tsx
kherP: updated the value using the list

develop/components/pages/TooltipPage.tsx
```diff
@@ -30,33 +89,108 @@ const TooltipPage: React.FunctionComponent = () => {
...
+                                condensed
+                            />
+                        </div>
+                        <div className="col">
+                            <p>Triggers</p>
+                            <RadioGroup
+                                name="trigger"
+                                list={triggerList}
+                                value={trigger}
+                                onChange={(e) => setTrigger(e.currentTarget.value as TooltipTrigger)}
```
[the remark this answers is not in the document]
kherP: added event type for change event

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +12,92 @@ export interface TooltipMessageGroupItem {
...
+
+    componentDidMount() {
+        const randID: string = this.constructId();
+        this.constructTooltipContentContainer(randID);
+    }
+
+    componentDidUpdate(prevProps: TooltipProps) {
+        if (prevProps.disableAutoPosition !== this.props.disableAutoPosition) {
+            if (!!this.state.tooltipPositionChecker) {
+                this.state.tooltipPositionChecker.toggleAutoPlacement(this.props.disableAutoPosition);
```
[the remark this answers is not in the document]
kherP: removed tooltipPositionChecker from the state

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+            this.getWithinViewportPosition();
+        }
+    }
+
+    /**
+     * toggle tooltip
+     * @param toggle boolean
+     * @param e event triggering the changes
+     */
+    private onTooltipToggle = (e?: React.MouseEvent<HTMLDivElement> | React.FocusEvent<HTMLDivElement>, toggle?: boolean) => {
```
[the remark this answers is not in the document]
kherP: added void

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    /**
+     * retrieve position within viewport
+     * @param position selected position
+     */
+    private getPositionWithinViewport(position: TooltipPosition): void {
+        const selectedPlacement: TooltipPlacement = tooltipPlacements[position];
+        if (selectedPlacement && selectedPlacement.relatedPoints.every((point) => this.isPointOverflow(point))) {
```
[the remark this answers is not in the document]
kherP: added type for point

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+        if (tooltipContainer) {
+            this.tooltip = tooltipContainer;
+        }
+    }
+
+    /**
+     * enable or disable auto placement for tooltip
+     * @param toggle boolean
+     */
+    toggleAutoPlacement(toggle: boolean) {
```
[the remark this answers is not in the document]
kherP: added return type

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+    constructor(refContainer: HTMLDivElement, disableAutoPlacement: boolean) {
+        this.reference = refContainer.querySelector(".tooltip-reference");
+        this.disableAutoPlacement = disableAutoPlacement;
+    }
+
+    /**
+     * add generated tooltip container
+     * @param tooltipContainer tooltip container
+     */
+    addTooltipContainer(tooltipContainer: HTMLDivElement) {
```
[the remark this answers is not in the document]
kherP: added return type

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+     * @param position selected position
+     */
+    private getPositionWithinViewport(position: TooltipPosition): void {
+        const selectedPlacement: TooltipPlacement = tooltipPlacements[position];
+        if (selectedPlacement && selectedPlacement.relatedPoints.every((point) => this.isPointOverflow(point))) {
+            this.getPlacement(position);
+        } else {
+            const possiblePlacements: Array<TooltipPosition> = ["top", "left", "right", "bottom", "bottom-left", "bottom-right", "left-bottom", "left-top", "right-bottom", "right-top", "top-left", "top-right"];
+            possiblePlacements.some((pos: TooltipPosition) => {
+                if (tooltipPlacements[pos].relatedPoints.every((point) => this.isPointOverflow(point))) {
```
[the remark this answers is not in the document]
kherP: added type for point

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+            case "bottom": return referenceRect.bottom + tooltipRect.height < window.innerHeight;
+            default: return referenceRect.top - tooltipRect.height > 0;
+        }
+    }
+
+    /**
+     * get current placement of tooltip
+     * @param pos selected position
+     */
+    private getPlacement(pos: TooltipPosition): void {
```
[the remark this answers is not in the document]
kherP: renaming all getXX method to setXX if return is void

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+        const newStyle = this.getNewStyle(pos);
```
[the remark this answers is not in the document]
kherP: added type

develop/components/pages/TooltipPage.tsx
omarVodiaK: what about an alert instead of the log, there is no guaranties the use will have the console opened.

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +12,92 @@ export interface TooltipMessageGroupItem {
...
+            classNames[1] = this.props.theme;
+            this.state.tooltipContainer.className = classNames.join(" ");
+        }
+        if (!!this.state.tooltipContainer && (prevProps.content !== this.props.content || prevProps.message !== this.props.message || prevProps.messageGroup !== this.props.messageGroup)) {
+            this.setTooltipContent(this.state.tooltipContainer);
+        }
+
+        if (!!this.props.message || !!this.props.messageGroup || !!this.props.onClick || !!this.props.title || !!this.props.customSvg || !!this.props.triggerOnHover || !!this.props.width) {
+            console.warn(
+                "%cmessage, messageGroup, onClick, title, customSvg, triggerOnHover, and width %cattributes will be %cdeprecated soon.",
```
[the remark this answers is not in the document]
kherP: changed to normal console.warn

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
         );
     }
+
+    /**
+     * construct tooltip content container and append to root tooltip container
+     * @param tooltipId unique tooltip id
+     */
+    private constructTooltipContentContainer = (tooltipId: string): void => {
+        const tooltipRootArr: HTMLCollectionOf<Element> = document.body.getElementsByClassName(this.tooltipRootClassName);
+        let tooltipRootRef: Element = tooltipRootArr && tooltipRootArr[0];
```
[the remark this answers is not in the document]
kherP: removed and replaced by createPortal

src/Tooltip/Tooltip.tsx
kherP: removed

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+    private setTooltipContent = (tooltip: HTMLDivElement): void => {
+        let content: string | React.ReactNode = this.props.content;
+        if (!this.props.content) {
+            if (this.props.message) {
+                content = <TooltipMessage {...this.props} />;
+            } else if (this.props.messageGroup) {
+                content = <TooltipMessageGroup {...this.props} />;
+            }
+        }
+        ReactDOM.render(<TooltipContent content={content} />, tooltip);
```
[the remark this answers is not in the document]
kherP: removed

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+    }
+
+    /**
+     * set tooltip visibility
+     * @param isVisible boolean
+     */
+    private setTooltipVisibility = (isVisible: boolean): void => {
+        this.state.tooltipContainer.style.display = isVisible ? "block" : "none";
```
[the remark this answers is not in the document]
kherP: removed

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +112,162 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+     * toggle tooltip
+     * @param toggle boolean
+     * @param e event triggering the changes
+     */
+    private onTooltipToggle = (e?: React.MouseEvent<HTMLDivElement> | React.FocusEvent<HTMLDivElement>, toggle?: boolean) => {
+        const isVisible: boolean = toggle !== undefined ? toggle : !this.state.visible;
+        this.setState({
+            visible: isVisible
+        });
+        this.setTooltipVisibility(isVisible);
```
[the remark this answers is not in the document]
kherP: removed

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+                    return true;
+                }
+            });
+        }
+    }
+
+    /**
+     * retrieve new tooltip style
+     * @param position position selected
+     */
```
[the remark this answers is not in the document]
kherP: added return type

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * get current placement of tooltip
+     * @param pos selected position
+     */
+    private getPlacement(pos: TooltipPosition): void {
+        const newStyle = this.getNewStyle(pos);
+        if (newStyle) {
+            Object.keys(newStyle).map((key: string) => {
+                this.tooltip.style[key] = newStyle[key];
```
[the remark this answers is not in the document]
kherP: removed

src/Tooltip/placement.ts
```diff
@@ -0,0 +1,262 @@
...
+            });
+        }
+        const classNames: Array<string> = this.tooltip.className.split(" ");
+        classNames[classNames.length - 1] = pos;
+        this.tooltip.className = classNames.join(" ");
```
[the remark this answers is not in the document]
kherP: removed

develop/components/pages/TooltipPage.tsx
kherP: removed the example as we are deprecating the function

develop/components/pages/TooltipPage.tsx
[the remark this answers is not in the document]
yousifalraheem: The documentation should not through console logs that are meant for the user. Since it won't strike the user to check the console for output messages. After we add a toast component, maybe we can consider showing these messages in a toast notification.

develop/components/pages/TooltipPage.tsx
```diff
@@ -1,22 +1,51 @@
...
+        { label: "Purple", value: "purple" },
+        { label: "Success", value: "success" },
+        { label: "Warning", value: "warning" },
+    ];
+    const positionList: Array<RadioListModel<TooltipPosition>> = [
+        { label: "Top", value: "top" },
+        { label: "Top Left", value: "top-left" },
+        { label: "Top Right", value: "top-right" },
+        { label: "Right", value: "right" },
+        { label: "Right Top", value: "right-top" },
```
omarVodiaK: wondering what is the difference between this line and { label: "Top Right", value: "top-right" }

yousifalraheem: Top right will show the tooltip on top but the content is on the right. This used to be helpful when you have a lot of content inside but not enough space. But with our new approach of using auto positioning, this might get deprecated.

yousifalraheem: The second direction in the `top-right` is for the content

src/Tooltip/Tooltip.tsx
yousifalraheem: I was talking about the naming, don’t call it icon. The classname and the prop name are both called icon

develop/components/pages/TooltipPage.tsx
```diff
@@ -30,33 +89,108 @@ const TooltipPage: React.FunctionComponent = () => {
...
-                            id="dismissable-tooltip"
+                            content={content === 0 && (hasNodeAsContent ? nodeTooltipContent : defaultTooltipContent)}
+                            message={content === 1 && defaultTooltipContent}
+                            messageGroup={content === 2 && defaultMessageGroup}
+                            customSvg={customSvg && moneySVG}
+                            position={position}
+                            theme={theme}
+                            trigger={trigger}
+                            disableAutoPosition={disableAutoPosition}
+                            onVisibleChange={hasCallback && (() => console.log("callback"))}
```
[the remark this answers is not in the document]
omarVodiaK: Use alert instead of console.log

src/Tooltip/Tooltip.tsx
kherP: rename classname to default-content

develop/components/pages/TooltipPage.tsx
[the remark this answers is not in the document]
kherP: use notification instead of console.log

src/Tooltip/Tooltip.tsx
```diff
@@ -53,68 +83,125 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
-                    onMouseLeave={() => this.props.triggerOnHover && this.toggleTooltip(false)}
+                    id={this.state.referenceId}
+                    ref={this.containerRef}
+                    className="tooltip-reference"
+                    tabIndex={-1}
+                    onClick={(e: React.MouseEvent<HTMLDivElement>) => ((!this.props.trigger && !this.props.triggerOnHover) || this.props.trigger === "click") && this.onTooltipToggle(e)}
+                    onMouseEnter={(e: React.MouseEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, true)}
+                    onMouseLeave={(e: React.MouseEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, false)}
+                    onTouchStart={(e: React.TouchEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, true)}
+                    onTouchEnd={(e: React.TouchEvent<HTMLDivElement>) => (this.props.triggerOnHover || this.props.trigger === "hover") && this.onTooltipToggle(e, false)}
```
DanSallau: Avoid arrow functions inside the render cycle of class component , especially when they are many. read [here](https://www.freecodecamp.org/news/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36/)

So instead, try

`onClick={this.handleClick}` 

```
handleClick(e: React.MouseEvent<HTMLDivElement>) {
((!this.props.trigger && !this.props.triggerOnHover) || this.props.trigger === "click") && this.onTooltipToggle(e);
}
```

src/__utils/Overlay/Overlay.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * on scroll callback
+     */
+    const onScroll = React.useCallback((ev: Event): void => {
+        const target: HTMLDivElement = ev.target as HTMLDivElement;
+        if (props.show && target.contains(props.overlayReference())) {
+            setPlacementWithCoords(tooltipPositionChecker.getPosition(props.position || "top"));
+        }
+    }, [props.show]);
```
DanSallau: missing `props.position`  in the array.

kherP: added

src/Tooltip/Tooltip.tsx
kherP: updated

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +11,64 @@ export interface TooltipMessageGroupItem {
...
 export class Tooltip extends React.Component<TooltipProps, TooltipState> {
+    private containerRef: React.RefObject<HTMLDivElement> = React.createRef();
+
     constructor(props: TooltipProps) {
         super(props);
 
         this.state = {
-            toggle: false
+            visible: false,
+            referenceId: "",
```
mario-subo: Any particular reason why the referenceId is being set on componentDidMount and not here?

src/Tooltip/Tooltip.tsx
```diff
@@ -8,37 +11,64 @@ export interface TooltipMessageGroupItem {
...
     width?: number;
+    content?: string | React.ReactNode;
+    trigger?: TooltipTrigger;
+    disableAutoPosition?: boolean;
+    onVisibleChange?: (event: React.MouseEvent<HTMLDivElement> | React.FocusEvent<HTMLElement> | React.TouchEvent<HTMLDivElement>, visible: boolean) => void;
+}
+interface TooltipState {
+    visible: boolean;
+    referenceId: string;
+    tooltipContainer: HTMLDivElement;
```
mario-subo: Where is this being used?

kherP: removed the unused attribute

src/Tooltip/Tooltip.tsx
kherP: removed didmount, inititate in state

src/Tooltip/Tooltip.tsx
```diff
@@ -8,113 +11,198 @@ export interface TooltipMessageGroupItem {
...
+export type TooltipTrigger = "hover" | "click" | "focus";
 export type TooltipTheme = "default" | "light" | "primary" | "warning" | "success" | "danger" | "purple";
-export type TooltipPosition = "top" | "bottom" | "left" | "right" | "top-right" | "top-left" | "bottom-right" | "bottom-left" | "left-top" | "left-bottom" | "right-top" | "right-bottom";
+export type TooltipPosition = ElementPosition;
 
 export interface TooltipProps {
     className?: string;
     customSvg?: React.ReactNode;
     id?: string;
+    ref?: React.LegacyRef<Tooltip>;
```
mario-subo: has this been removed yet?

mario-subo: Also don't forget to remove from the README file as well if it mentions ref

kherP: this ref is needed for forceShow function

develop/components/pages/TooltipPage.tsx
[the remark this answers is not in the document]
mario-subo: has this been resolved @kherP ?

PR #113, opened 2020-01-30
dropdown alert style fix

### LINKED ISSUE #106: importing Dropdown component overrides some default bootstrap styles
**Describe the bug**
The dropdown component incorrectly overrides a bootstrap class

**To Reproduce**
1. Create a new project and add SEB bootstrap
2. Display a basic alert using the `alert alert-danger` class name
3. add react-components to your project and import a dropdown component
4. The alert style will get overriden by the dropdown component css rules

**Expected behavior**
components should not override default bootstrap classes like `alert`

mario-subo:
- [x] Added a toggle which shows and hides example error messages under each dropdown example
image

Oh and also fixes #106 👍

### DIFF
--- a/develop/components/pages/DropdownPage.tsx
+++ b/develop/components/pages/DropdownPage.tsx
@@ -11,7 +11,7 @@ const DropdownPage: React.FunctionComponent = () => {
     const [dropDownList4Selected, setDropdownList4Selected] = React.useState<DropdownItem>(null);
     const [dropDownList5Selected, setDropdownList5Selected] = React.useState<Array<DropdownItem>>([]);
     const [disabled, setDisabled] = React.useState<boolean>(false);
-    const [error, setError] = React.useState<string>(null);
+    const [error, setError] = React.useState<string>("");
 
     const handleToggleError = React.useCallback(() => {
         setError((currentError) => (currentError === null) ? "Example error message" : null);
--- a/develop/components/pages/DropdownPage.tsx
+++ b/develop/components/pages/DropdownPage.tsx
@@ -11,6 +11,11 @@ const DropdownPage: React.FunctionComponent = () => {
     const [dropDownList4Selected, setDropdownList4Selected] = React.useState<DropdownItem>(null);
     const [dropDownList5Selected, setDropdownList5Selected] = React.useState<Array<DropdownItem>>([]);
     const [disabled, setDisabled] = React.useState<boolean>(false);
+    const [error, setError] = React.useState<string>(null);
+
+    const handleToggleError = React.useCallback(() => {
+        setError((currentError) => (currentError === null) ? "Example error message" : null);
+    }, [setError]);
 
     return (
         <div className="route-template container">
@@ -32,6 +37,7 @@ const DropdownPage: React.FunctionComponent = () => {
                             selectedValue={dropDownList1Selected}
                             onChange={(value: DropdownItem) => setDropdownList1Selected(value)}
                             disabled={disabled}
+                            error={error}
                         />
                     </div>
 
@@ -47,6 +53,7 @@ const DropdownPage: React.FunctionComponent = () => {
                             placeholder="Multi option"
                             multi={true}
                             disabled={disabled}
+                            error={error}
                         />
                     </div>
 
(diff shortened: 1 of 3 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -322,11 +322,11 @@ const Dropdown: React.FunctionComponent<DropdownProps> = (props: DropdownProps):
     // Display the custom dropdown with native elements if prop is set to native
     if (props.native) {
         return (
-            <>
+            <div className="custom-native-dropdown">
```
yousifalraheem: Is this issue only found in the **native** variant?

mario-subo: No, both variants previously used the `alert alert-danger` class names only.

develop/components/pages/DropdownPage.tsx
```diff
@@ -11,6 +11,11 @@ const DropdownPage: React.FunctionComponent = () => {
     const [dropDownList4Selected, setDropdownList4Selected] = React.useState<DropdownItem>(null);
     const [dropDownList5Selected, setDropdownList5Selected] = React.useState<Array<DropdownItem>>([]);
     const [disabled, setDisabled] = React.useState<boolean>(false);
+    const [error, setError] = React.useState<string>(null);
```
omarVodiaK:
```suggestion
    const [error, setError] = React.useState<string>("");
```
To avoid showing null in the error field in some edge cases.

develop/components/pages/DropdownPage.tsx
```diff
@@ -32,6 +37,7 @@ const DropdownPage: React.FunctionComponent = () => {
                             selectedValue={dropDownList1Selected}
                             onChange={(value: DropdownItem) => setDropdownList1Selected(value)}
                             disabled={disabled}
+                            error={error}
```
omarVodiaK: Putting an error message on all dropdowns can be confusing for the user, i suggest to have one dropdown with the error instead

src/Dropdown/dropdown-style.scss
```diff
@@ -295,6 +280,24 @@ $white: #fff;
     }
 }
 
+// Native version
+.custom-native-dropdown {
+    margin-bottom: 8px;
```
omarVodiaK: Is this margin to leave a space between the native dropdown and the error div ?
If yes i suggest to have the margin set on the error div instead, that way the margin will only be added if there is an error, otherwise the dropdown will be displayed normally

develop/components/pages/DropdownPage.tsx
mario-subo: All the errors will be shown only if you toggle it on the bottom of the page;. Same as disabled

src/Dropdown/dropdown-style.scss
mario-subo: No this is margin is a default margin for the entire dropdown, error or not, native or regular. the `.custom-dropdown` has the same margin-bottom

develop/components/pages/DropdownPage.tsx
```diff
@@ -11,6 +11,11 @@ const DropdownPage: React.FunctionComponent = () => {
     const [dropDownList4Selected, setDropdownList4Selected] = React.useState<DropdownItem>(null);
     const [dropDownList5Selected, setDropdownList5Selected] = React.useState<Array<DropdownItem>>([]);
     const [disabled, setDisabled] = React.useState<boolean>(false);
+    const [error, setError] = React.useState<string>("");
+
+    const handleToggleError = React.useCallback(() => {
+        setError((currentError) => (currentError === null) ? "Example error message" : null);
```
DanSallau: What is this `currentError` and where is it coming from ?

develop/components/pages/DropdownPage.tsx
mario-subo: @omarVodiaK can I resolve this of you want me to make changes?

develop/components/pages/DropdownPage.tsx
mario-subo: The React useState setter (setError in thius case) accepts a new value **or alternatively** a function which provides the current state and lets you return a new state.

mario-subo: https://reactjs.org/docs/hooks-reference.html#functional-updates

PR #114, opened 2020-02-03
refactor(button): updates button theme names

### LINKED ISSUE #110: Change name of Alternative button to Button link
### Change
The button **Button link** used to be called Alternative button.
image
### Reason
The name Button link says what it is.

See the full specification here: 
https://designlibrary.sebgroup.com/components/forms/button/

Feel free to update this in your frameworks so that we use the same language on components. If this is not relevant to you, please close the issue with a comment. Thank you!

/Ulrika,
Design Management team

mario-subo:
Updates the theme names: 
- Alternative -> Secondary,
- Secondary -> Primary outlined,
- Anchor -> Link

Since there are some dependencies between design library and Bootstrap, I decided to just follow the naming conventions from [SEB Bootstrap](https://sebgroup.github.io/bootstrap/buttons). 

fix #110

image

### DIFF
--- a/src/Button/Button.test.tsx
+++ b/src/Button/Button.test.tsx
@@ -44,11 +44,12 @@ describe("Component: Button", () => {
     describe("Should render supported themes", () => {
         const list: Array<ButtonTestItem<ButtonTheme, string>> = [
             { value: "primary", expected: "btn-primary" },
-            { value: "secondary", expected: "btn-outline-primary" },
-            { value: "alternative", expected: "btn-secondary" },
+            { value: "outline-primary", expected: "btn-outline-primary" },
+            { value: "secondary", expected: "btn-secondary" },
             { value: "ghost-dark", expected: "btn-ghost-dark" },
             { value: "ghost-light", expected: "btn-ghost-light" },
-            { value: "anchor", expected: "btn-anchor" },
+            { value: "anchor", expected: "btn-link" },
+            { value: "link", expected: "btn-link" },
             { value: "danger", expected: "btn-danger" },
             { value: "unsupported-theme" as any, expected: "btn-primary" },
         ];
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -27,7 +27,7 @@ export const Button: React.NamedExoticComponent<ButtonProps> = React.memo((props
 
     function getButtonTheme(): string {
         switch (props.theme) {
-            case "anchor": return "link";
+            case "anchor": return "link"; @deprecated
             case "outline-primary": case "secondary": case "ghost-dark": case "ghost-light": case "danger": case "link": return props.theme;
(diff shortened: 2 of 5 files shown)


src/Button/Button.test.tsx
```diff
@@ -44,11 +44,12 @@ describe("Component: Button", () => {
...
         const list: Array<ButtonTestItem<ButtonTheme, string>> = [
             { value: "primary", expected: "btn-primary" },
-            { value: "secondary", expected: "btn-outline-primary" },
-            { value: "alternative", expected: "btn-secondary" },
+            { value: "outline-primary", expected: "btn-outline-primary" },
+            { value: "secondary", expected: "btn-secondary" },
             { value: "ghost-dark", expected: "btn-ghost-dark" },
             { value: "ghost-light", expected: "btn-ghost-light" },
-            { value: "anchor", expected: "btn-anchor" },
+            { value: "anchor", expected: "btn-link" },
```
omarVodiaK: i can't see anchor anymore in the bootstrap/buttons, maybe we should remove it from here as well ?

src/Button/Button.tsx
```diff
@@ -27,9 +27,8 @@ export const Button: React.NamedExoticComponent<ButtonProps> = React.memo((props
 
     function getButtonTheme(): string {
         switch (props.theme) {
-            case "secondary": return "outline-primary";
-            case "alternative": return "secondary";
-            case "primary": case "ghost-dark": case "ghost-light": case "anchor": case "danger": return props.theme;
+            case "anchor": return "link";
```
omarVodiaK: we should remove anchor from here as well

src/Button/Button.test.tsx
yousifalraheem: We should only keep it so that we don't break the styles of people who are already using it. We should mention it in the release notes that it is deprecated and will be removed soon. Give them a change to switch to `link` before we remove it.

omarVodiaK: Yes understand, can we add @deprecated to it so that we keep track of it?

src/Button/Button.tsx
omarVodiaK:
```suggestion
            case "anchor": return "link"; @deprecated
```

mario-subo: I added a console.warn instead, because "anchor" is a string literal option, there's no good way to put a @deprecated mark on it. If you know a better way let me know. See commit

src/Button/Button.test.tsx
DanSallau: shouldn't the value here be `link` ?

yousifalraheem: We are deprecating `anchor`, but we so support it. Therefore, this tests of passing the old theme yields the same result as using the new one.

PR #134, opened 2020-02-20
refactor(table): move and reuse table component

DanSallau:


src/Table/Table.tsx
```diff
@@ -227,6 +227,118 @@ const ActionColumn: React.FunctionComponent<ActionColumnProps> = (props: ActionC
...
+    row: TableRow;
+    rowsAreCollapsable: boolean;
+    tableRef: React.RefObject<HTMLTableElement>;
+    type: RowTypes;
+    useRowCollapse: boolean;
+    useRowSelection: boolean;
+    useShowActionColumn: boolean;
+}
+
+const RowUI: React.FunctionComponent<RowUIProps> = (props: RowUIProps) => {
```
mario-subo: Could you name this `TableRow`? I think it's more descriptive.

DanSallau: We have an interface TableRow already, which is an interface for tableRow. I call this UI because it's UI. Also you notice we have another UI tableUI, which is the main UI for the table. So basically UI here is my way of categorizing them.

PR #153, opened 2020-03-05
feat(table-component): add filter feature for table

DanSallau:


src/Table/Table.tsx
```diff
@@ -3,12 +3,64 @@ import { randomId } from "../__utils/randomId";
...
+        />
+    </svg>
+);
+const timesIcon: JSX.Element = (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512">
+        <path d="M217.5 256l137.2-137.2c4.7-4.7 4.7-12.3 0-17l-8.5-8.5c-4.7-4.7-12.3-4.7-17 0L192 230.5 54.8 93.4c-4.7-4.7-12.3-4.7-17 0l-8.5 8.5c-4.7 4.7-4.7 12.3 0 17L166.5 256 29.4 393.2c-4.7 4.7-4.7 12.3 0 17l8.5 8.5c4.7 4.7 12.3 4.7 17 0L192 281.5l137.2 137.2c4.7 4.7 12.3 4.7 17 0l8.5-8.5c4.7-4.7 4.7-12.3 0-17L217.5 256z" />
+    </svg>
+);
 
 export type DataItem = { [name: string]: any } & TableRow;
```
kherP: suggested change:  `export type DataItem<T> = T & TableRow;`

yousifalraheem: You can make it optional
```suggestion
export type DataItem<T = { [name: string]: any }> = T & TableRow;
```

src/Table/Table.tsx
```diff
@@ -3,14 +3,66 @@ import { randomId } from "../__utils/randomId";
...
+);
+const ellipsis: JSX.Element = (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512">
+        <path d="M192 256c0 17.7-14.3 32-32 32s-32-14.3-32-32 14.3-32 32-32 32 14.3 32 32zm88-32c-17.7 0-32 14.3-32 32s14.3 32 32 32 32-14.3 32-32-14.3-32-32-32zm-240 0c-17.7 0-32 14.3-32 32s14.3 32 32 32 32-14.3 32-32-14.3-32-32-32z" />
+    </svg>
+);
+const sortAsc: JSX.Element = (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 644">
+        <path
+            fill="#adadad"
```
yousifalraheem: I just noticed, why the colors are hardcoded in here? What if the user chooses to change it?

src/Table/Table.tsx
```diff
@@ -179,8 +239,8 @@ const ActionColumn: React.FunctionComponent<ActionColumnProps> = (props: ActionC
                 >
                     {props.primaryActionButton.label}
                 </button>
-            }
-            {props.actionLinks && props.actionLinks.length &&
+            )}
+            {props.actionLinks && props.actionLinks?.length && (
```
yousifalraheem: I think if you test this with an empty action links array it will actually render `0` here. It's better use:
```tsx
{props.actionLinks && props.actionLinks?.length ? (
    /* your code */
) : null }
```

src/Table/Table.tsx
DanSallau: initially, we use two separate angle top and down for the sorting ASC and DESC. But then again, seb bootstrap/design library has a icons for both ASC and Desc and default that fit their desc. Also theirs come with the whole package, meaning ASC icon, comes with the two angle up and down together , with the angle top section set as active. The same applies for DESC. Unless you can find a way to fill a certain section of the icon/svg leaving the rest, it will not be as dynamic.

src/Table/Table.tsx
```diff
@@ -521,483 +617,540 @@ interface TableProps {
...
+            },
+            [tableRows, currentTableRows, props.onRowExpanded]
+        );
+        // functions -----------------------------------------------------------------------------
+        /**
+         *
+         * @param rows The table or or data to initialize rows from
+         */
+        const getRows = React.useCallback((rows: Array<DataItem<any>>): Array<TableRow> => {
+            const updatedRows: Array<TableRow> = rows.map((row: TableRow, index: number) => {
```
kherP: `rows?.map` else the code will break when data is set to null

src/Table/Table.tsx
```diff
@@ -521,483 +617,540 @@ interface TableProps {
...
+            }
+        }, [props.currentpage, props.offset, tableRows]);
 
-    const rowsAreCollapsable = React.useCallback((): boolean => {
-        return currentTableRows.some((row: TableRow) => {
+        const rowsAreCollapsable = React.useCallback((): boolean => {
             return (
-                ((row.subRows.length > 0) || row.rowContentDetail) ||
-                (row.subRows.some((subRow: TableRow) => subRow.rowContentDetail || (row.subRows && row.subRows.length > 0)))
+                currentTableRows.some((row: TableRow) => {
```
kherP: `currentTableRows?.map` else table will break when data is set to null

src/Table/Table.tsx
```diff
@@ -521,483 +617,540 @@ interface TableProps {
...
+                }
             }
-        }
-    }, [props.searchProps?.searchInColumns, props.searchProps?.searchText]);
 
-    React.useEffect(() => {
-        if (!!props.searchProps?.onSearch) {
-            if (props.searchProps?.triggerSearchOn === "Submit") {
-                doSearch();
+            const actionColumnIsOpened: boolean = tableRows.some((row: TableRow) => {
```
kherP: `tableRows?.map` else it will break when data is set to null

src/Table/Table.tsx
```diff
@@ -521,483 +617,540 @@ interface TableProps {
...
+        }, [tableColumns]);
 
-        const actionColumnIsOpened: boolean = tableRows.some((row: TableRow) => {
-            return row.actionsDropdownDropped || row.subRows?.some((sub: TableRow) => sub.actionsDropdownDropped);
-        });
-        setAnOpenedAction(actionColumnIsOpened);
-    }, [currentTableRows]);
+        React.useEffect(() => {
+            if (!!props.onRowSelected) {
+                const notAllsAreRowsSelected: boolean = tableRows.some((row: TableRow) => !row.selected);
```
kherP: `tableRows?.some` as well

src/Table/Table.tsx
DanSallau: You can't have all these without a row hence the reason why i am not checking .

src/Table/Table.tsx
DanSallau: I think it makes sense here.

src/Table/Table.tsx
DanSallau: I have added some checks on the rows across the project to be  safer.

src/Table/Table.tsx
yousifalraheem: Just remember you put a default value in return since `rows?.map` might return null in case there are no rows. Since the logic of the component relies on the existence of an array it might crash here.

Something like this:
```typescript
rows?.map(() => { /** your logic */ }) || []
```

src/Table/Table.tsx
yousifalraheem: @DanSallau please check this [PR](https://github.com/sebgroup/react-components/pull/165) for reference

PR #154, opened 2020-03-05
Refactor/remove lodash add buffer

DanSallau:


package.json
```diff
@@ -44,10 +44,10 @@
 	},
 	"dependencies": {
 		"@sebgroup/bootstrap": "^5.1.0",
+		"buffer": "^5.4.3",
```
yousifalraheem: where do you use this? I couldn't find the implementation

develop/__utils/enum.util.ts
```diff
@@ -0,0 +1,4 @@
+export const enum sortDirectionTypes {
```
yousifalraheem: why is this a separate file?

yousifalraheem:
Can you elaborate on why are you replacing lodash with a custom utility?

yousifalraheem:
@DanSallau Is this PR still valid?

develop/__utils/enum.util.ts
DanSallau: Because we don't have an enum util file of course . It is an an enum which supposedly belong in an enum file hence the reason why it's separate.

PR #164, opened 2020-03-18
Chore/add slider alternative style

mario-subo:
captured


EldRoos:
Will check this later today. Good work.

EldRoos:
Found one thing, missing round corners on slider track.

mario-subo:
> Found one thing, missing round corners on slider track.

The change I did adds an optional alternative thumb style only and it doesn't change the existing look of the current design library spec. It wouldn't make sense to change the slider track as that would be an overall style change which would affect every thumb style. So since it's not (just like the smaller thumb) part of design library yet we can't make that change yet. I urge you to help out the design library in streamlining and updating this component if you think it's needed. Thanks 

For reference: http://designlibrary.sebank.se/components/slider/

EldRoos:
Nope you already make the slider track 1 pixel thinner. So nope it's not a thumb only change allready. 

Also why have this change at all if we just add half of it.

EldRoos:
Henrik Hoglund is again working on the new sketches. Will check status with him later on.

src/Slider/Slider.tsx
```diff
@@ -2,10 +2,19 @@ import * as React from "react";
...
+    </svg>
+);
+const angleRightIcon: JSX.Element = (
+    <svg name="angle-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 192 512">
+        <path d="M166.9 264.5l-117.8 116c-4.7 4.7-12.3 4.7-17 0l-7.1-7.1c-4.7-4.7-4.7-12.3 0-17L127.3 256 25.1 155.6c-4.7-4.7-4.7-12.3 0-17l7.1-7.1c4.7-4.7 12.3-4.7 17 0l117.8 116c4.6 4.7 4.6 12.3-.1 17z" />
+    </svg>
+);
 
 export type SliderTheme = "primary" | "inverted" | "success" | "danger" | "warning" | "purple";
+export type SliderAppearance = "normal" | "smaller";
```
omarVodiaK, line 17: SEB are using value like:
```css
$font-weight-lighter:         lighter !default;
$font-weight-light:           300 !default;
$font-weight-normal:          400 !default;
$font-weight-bold:            700 !default;
$font-weight-bolder:          bolder !default;
```
i suggest to use small instead of smaller, since smaller can be introduced later
```javascript
export type SliderAppearance = "normal" | "small";
```

yousifalraheem:
@mario-subo It's time to bring this PR back to life. We have a new design. Please see this issue #298

src/Slider/Slider.tsx
```diff
@@ -36,18 +37,31 @@ export interface SliderProps {
     showTicks?: boolean;
     step?: number;
     theme?: SliderTheme;
+    appearance?: SliderAppearance;
```
yousifalraheem, line 40: I don't like this approach, let's just follow the way we used to handle this which is through passing a boolean value of alternative which will be like this:
```tsx
<Slider value={value} alternative />
```
The reason is, there are no other appearances, there is just a default and another one. Comparing this to HTML elements, the disabled state is set similarly.

src/Slider/Slider.tsx
```diff
@@ -111,29 +125,31 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
...
                 style.transform = "rotateY(180deg)";
             } else {
                 zeroPosition = size ? Math.abs(100 - (max / size) * 100) : 0;
-                style.left = `calc(${zeroPosition}% + 27px)`;
+                style.left = `calc(${zeroPosition}% + ${width})`;
                 style.width = calculatedThumbPosition - zeroPosition + "%";
             }
         }
-        return style;
+        return { ...style };
```
yousifalraheem, line 152: This is unnecessary processing, you don't need to spread something that you just created.

src/Slider/Slider.tsx
omarVodiaK: I agree, the objective is to change the UI without having to change code or break it.
```typescript 
<Slider value={value} />
```
should work fine
If you really insist on keeping the appearance then please add a default value to it
```typescript
appearance?: SliderAppearance = SliderAppearance.normal;
```

src/Slider/Slider.tsx
```diff
@@ -165,7 +181,7 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
     return (
         <div className={"form-group custom-slider" + (props.className ? ` ${props.className}` : "") + (props.disabled ? " disabled" : "")}>
             {props.label && <label className="custom-label">{props.label}</label>}
-            <div className={"input-field" + (props.labels && props.labels.length ? " has-labels" : "")}>
+            <div className={"input-field" + (props.labels && props.labels.length ? " has-labels" : "") + (props.appearance ? ` ${props.appearance}` : ` ${defaultAppearance}`)}>
```
omarVodiaK, line 184: classname dependecy please :D

src/Slider/slider-style.scss
```diff
@@ -6,13 +6,10 @@ $disabled-selected: $gray-500;
...
-
-@mixin transition($props) {
-    -webkit-transition: $props;
-    -moz-transition: $props;
-    -o-transition: $props;
-    transition: $props;
-}
+$slider-width-normal: 24px;
+$slider-height-normal: $slider-width-normal;
+$slider-width-alternative: 56px;
```
omarVodiaK, line 11: 3.5rem

src/Slider/slider-style.scss
```diff
@@ -6,13 +6,10 @@ $disabled-selected: $gray-500;
...
 $disabled-thumb-bg: #9a9a9a;
 $disabled-thumb-border-color: #8c8c8c;
-
-@mixin transition($props) {
-    -webkit-transition: $props;
-    -moz-transition: $props;
-    -o-transition: $props;
-    transition: $props;
-}
+$slider-width-normal: 24px;
```
omarVodiaK, line 9: 1.5 rem

src/Slider/slider-style.scss
```diff
@@ -6,13 +6,10 @@ $disabled-selected: $gray-500;
...
+$slider-height-normal: $slider-width-normal;
+$slider-width-alternative: 56px;
+$slider-height-alternative: 32px;
```
omarVodiaK, line 12: 2rem

src/Slider/Slider.tsx
mario-subo: Omar I added the default value here https://github.com/sebgroup/react-components/pull/164/files#diff-3080399ac815fe720993abb6fa017bf1R64

mario-subo: Yousif what about if we decide to add more appearances in the future? You approach only covers alternative and normal

src/Slider/Slider.tsx
mario-subo: good catch, you're right

src/Slider/slider-style.scss
mario-subo: then I should change all the sized in this file to rem. I can't change just this one

src/Slider/Slider.tsx
mario-subo: > I agree, the objective is to change the UI without having to change code or break it.
> 
> ```ts
> <Slider value={value} />
> ```
> 
> should work fine
> If you really insist on keeping the appearance then please add a default value to it
> 
> ```ts
> appearance?: SliderAppearance = SliderAppearance.normal;
> ```

The code won't break if you don't pass appearance. Yousif is just saying instead of appearance string we should just have alternative as boolean.

src/Slider/Slider.tsx
mario-subo: this is angular though so we can just se ngClass lol

src/Slider/slider-style.scss
mario-subo: > then I should change all the sized in this file to rem. I can't change just this one

src/Slider/slider-style.scss
mario-subo: > then I should change all the sized in this file to rem. I can't change just this one


so what's the verdict? Change all or keep all? @kherP @DanSallau @yousifalraheem ?

yousifalraheem:
@mario-subo 
The alternative slider looks more rounded than it used to before, check the previous version [here](https://sebgroup.github.io/react-components/#/slider)

src/Slider/Slider.tsx
```diff
@@ -36,18 +37,31 @@ export interface SliderProps {
...
     const [min, setMin] = React.useState<number>(props.min || 0);
     const [max, setMax] = React.useState<number>(props.max || 100);
     const [size, setSize] = React.useState<number>(0);
     const [labelsPositions, setLabelsPositions] = React.useState<Array<string>>([]);
     const [thumbPosition, setThumbPosition] = React.useState<number>(0);
     const [activeTrackStyles, setActiveTrackStyles] = React.useState<React.CSSProperties>({});
+    const appearanceSizesMap: AppearanceStyleMap = {
+        alternative: { width: "27px", offset: "56px" },
+        normal: { width: "5px", offset: "24px" },
+    };
```
kherP, line 63: maybe you can add react.usememo for this

src/Slider/Slider.tsx
```diff
@@ -111,29 +125,31 @@ const Slider: React.FunctionComponent<SliderProps> = (props: SliderProps): React
     function getActiveTrackStyles(): React.CSSProperties {
```
kherP, line 125: maybe can wrap this in useCallback

src/Slider/Slider.tsx
yousifalraheem: There will be no more appearances, that's for sure. In fact, it's more likely that we will deprecate the alternative appearance eventually. We're trying to get people to abide by Design Library standards, not help renegades.

src/Slider/Slider.tsx
yousifalraheem: I think you're confused, this is React🤣 .. you can leave it for now since the `classnames` dependency is happening on the refactoring branch. I will do it there when I get to it.

src/Slider/slider-style.scss
yousifalraheem: I don't think we need the slider to increase in size when the user increases the font size. Using pixels is fine in my opinion.

src/Slider/slider-style.scss
yousifalraheem: `rem` is meant to be used to have a dynamic size when using web readers. I don't think we need to increase the dimensions of a slider when the user increase the font size. It will blow it out of proportions.

src/Slider/Slider.tsx
yousifalraheem: `useMemo` is used to memorize the results of a function so you don't need to run it again, [read here](https://reactjs.org/docs/hooks-reference.html#usememo). However, if this is a constant that never changes, just drop it outside the function.

src/Slider/Slider.tsx
mario-subo: > this is angular though so we can just se ngClass lol

omg i don't know where i am anymore 🤣

PR #189, opened 2020-03-26
Feature/inline text box

DanSallau:
inline-text

### DIFF
--- a/develop/components/pages/TablePage.tsx
+++ b/develop/components/pages/TablePage.tsx
@@ -32,11 +32,13 @@ const TablePage: React.FunctionComponent = () => {
             {
                 label: "id",
                 accessor: "id",
-                canSort: false
+                canSort: false,
+                canEdit: false
             },
             {
                 label: "First Name",
-                accessor: "firstName"
+                accessor: "firstName",
+                canEdit: false
             },
             {
                 label: "Last Name",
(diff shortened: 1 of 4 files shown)


develop/components/pages/TablePage.tsx
```diff
@@ -206,6 +216,17 @@ const TablePage: React.FunctionComponent = () => {
                         <Table columns={columns} data={smallData} onRowSelected={(rows: Array<TableRow>) => {}} />
                     </div>
 
+                    <p>Here is an example with inline edit</p>
+                    <div className="result wide">
+                        <div className="row">
+                            <div className="col text-right">
+                                <Button title="Cancel" label="Cancel" disabled={!editMode} onClick={() => setEditMode("cancel")} className="mr-2" />
+                                <Button title="Update" label={editMode === "edit" ? "Save" : "Edit"} onClick={() => setEditMode(editMode === "edit" ? "save" : "edit")} />
```
omarVodiaK: label={editMode === "edit" ? "Save" : "Edit"} 
```suggestion
                                <Button title="Update" label={editMode === "edit" ? "Edit" : "Save"} onClick={() => setEditMode(editMode === "edit" ? "save" : "edit")} />
```

src/Table/Table.tsx
```diff
@@ -932,7 +1032,8 @@ export const Table: React.FunctionComponent<TableProps> = React.memo(
                     selected: row.selected || false,
                     actionsDropdownDropped: row.actionsDropdownDropped || false,
                     expanded: row.expanded || false,
-                    subRows: row.subRows ? getRows(row.subRows) : []
+                    subRows: row.subRows ? getRows(row.subRows) : [],
+                    isEditMode: row.isEditMode || false
```
kherP: why you need `|| false` here

DanSallau: Its initialisation, isEditMode is optional and just like the case is for the remaining objects in the data, we initialized all in getRows .

develop/components/pages/TablePage.tsx
DanSallau: No, the current one is correct omar. Edit comes before save, the default button text is edit

src/Table/Table.tsx
mario-subo: `!!row.expanded` <-- I prefer

PR #198, opened 2020-03-27
react node as cell

kherP:


src/Table/Table.tsx
```diff
@@ -43,6 +43,7 @@ export interface Column {
     label: string | React.ReactNode;
     accessor: string;
     canSort?: boolean;
+    cellTemplate?: (row: TableRow, cell: Cell) => React.ReactNode;
```
DanSallau: I don't see the need for this frankly speaking. Just Adding  a type reactNode to the value property should suffice. I don't think need a new property.

DanSallau:
If casting your react value to any allows you to render you reactElement, then all i will be expecting here are the change on the typings thereby allowing reactNode . I don't see the point of adding new property here .

src/Table/Table.tsx
kherP: to be honest, filtering by using another hidden columns is really just a workaround. I think we should keep this.

DanSallau: This PR address filtering or just rendering reactElement ?

develop/components/pages/TablePage.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { Table, Column, TableRow, PrimaryActionButton, ActionLinkItem, TableHeader, DataItem, FilterItem, FilterProps } from "../../../src/Table";
+import { Table, Column, TableRow, Cell, PrimaryActionButton, ActionLinkItem, TableHeader, DataItem, FilterItem, FilterProps } from "../../../src/Table";
```
yousifalraheem: You've imported `Cell` interface but never used it

src/Table/Table.tsx
```diff
@@ -51,7 +52,7 @@ export interface ActionLinkItem {
 }
 
 export interface PrimaryActionButton {
-    label: string;
+    label: string | React.ReactNode;
```
yousifalraheem: I believe `ReactNode` interface extends `ReactText` which is a `string` or `number`
```suggestion
    label: React.ReactNode;
```

src/Table/Table.tsx
yousifalraheem: I agree with @DanSallau here. It's not just here, we use the same approach will all our components. You only need a type change for `value` prop to `ReactNode` and that's should suffice

kherP: loop each `TableRow` then each `Cell` to set the value for each cell? Please reconsider.

PR #229, opened 2020-04-06
Resolve conflicts

yousifalraheem:


src/Dialogue/Dialogue.tsx
```diff
@@ -10,17 +10,17 @@ const TimesIcon: JSX.Element = (
...
-    disablePrimaryBtn?: boolean;
-    disableSecondaryBtn?: boolean;
     enableBackdropDismiss?: boolean;
     enableCloseButton?: boolean;
     header?: string | JSX.Element | React.ReactNode;
     id?: string;
     onDismiss?: (e?: React.MouseEvent<HTMLDivElement>) => void;
     primaryAction?: (e?: React.MouseEvent<HTMLButtonElement>) => void;
     primaryBtn?: string | JSX.Element;
+    primaryBtnDisabled?: boolean;
```
omarVodiaK, line 20: why is it in the past tense disablePrimaryBtn looks better

yousifalraheem: it used to be `disablePrimaryBtn` but this has been changed a while back when @mario-subo indicated that it sounds like an action instead of a state.

PR #256, opened 2020-04-10
Create gh-pages.yml

yousifalraheem:
This PR adds a workflow setup that triggers when code is pushed to **master** branch. The workflow with install dependencies, build the docs and publish them on `gh-pages` branch.

The settings of this repository are still not configured to retrieve docs from `gh-pages` yet, I will wait until the next push to verify it's integrity before changing the settings and adding `docs` folder to `.gitignore`


omarVodiaK:
Is there a token you need to generate to be able to do this ?

PR #278, opened 2020-04-12
fix(image-holder): cursor-pointer on clickable image

DanSallau:


src/Image/image-style.scss
```diff
@@ -2,4 +2,14 @@
...
+    &.link {
+        cursor: pointer;
+    }
+}
+
+.img-tag {
+    &.link {
+        cursor: pointer;
+    }
+}
```
yousifalraheem, line 15: Can we do this instead?
```suggestion
}

.img-tag,
.div-tag {
    &.link {
        cursor: pointer;
    }
}
```

src/Image/Image.tsx
```diff
@@ -22,7 +22,7 @@ export const Image: React.FunctionComponent<ImageProps> = React.memo(
                 {!props.useImgTag && (
                     <div
                         id={props.id}
-                        className={"div-tag" + (props.className ? ` ${props.className}` : "")}
+                        className={`div-tag${props.onClick ? " link" : ""} ${props.className ? props.className : ""}`}
```
yousifalraheem, line 25: This will leave an empty space if className is not passed. Here's how to fix it
```suggestion
                        className={`div-tag${props.onClick ? " link" : ""}${props.className ? ` ${props.className}` : ""}`}
```

src/Image/Image.tsx
```diff
@@ -37,7 +37,7 @@ export const Image: React.FunctionComponent<ImageProps> = React.memo(
                 {props.useImgTag && (
                     <img
                         id={props.id}
-                        className={"img-tag" + (props.className ? ` ${props.className}` : "")}
+                        className={`img-tag${props.onClick ? " link" : ""} ${props.className ? props.className : ""}`}
```
yousifalraheem, line 40: This will leave an empty space if className is not passed. Here's how to fix it
```suggestion
                        className={`img-tag${props.onClick ? " link" : ""}${props.className ? ` ${props.className}` : ""}`}
```

src/Image/Image.tsx
DanSallau: Oh, haven't given much thought about the className considering it's a preexisting prop. Nice catch.

DanSallau:
> @DanSallau
> It would be better if you create an internal state for `className`, you can calculate it with effects and pass it to either container

Is that really necessary for something that isn't gonna change or result in any local state effect?

yousifalraheem:
> > @DanSallau
> > It would be better if you create an internal state for `className`, you can calculate it with effects and pass it to either container
> 
> Is that really necessary for something that isn't gonna change or result in any local state effect?

It's just cleaner that repeating this statement twice:
```tsx
className={`img-tag${props.onClick ? " link" : ""}${props.className ? ` ${props.className}` : ""}`}
```

src/Image/Image.tsx
```diff
@@ -17,12 +17,33 @@ export interface ImageProps {
...
     (props: ImageProps): React.ReactElement<void> => {
+        const [imageClassName, setImageClassName] = React.useState<string>("");
+
+        React.useEffect(() => {
+            let className: string = "";
+            if (!props?.useImgTag) {
+                className += "div-tag";
+            } else {
+                className += "img-tag";
+            }
```
kherP, line 28: this can be simplified to `let className: string = props?.useImgTag ? "img-tag" : "div-tag";`

or reverse the if else condition so you don't need `!` in the first if condition

yousifalraheem, line 28:
```suggestion
            let className: string = props?.useImgTag ? "img-tag" : "div-tag"
```

DanSallau: LoL its the same with the render/UI. Kind of lazy i guess

kherP: extra `!` do extra unnecessary one bit work, haha, please change

PR #285, opened 2020-04-15
refactor(accordion): refactoring the accordion component

yousifalraheem:
the accordion component is refactored to heavily rely on bootstrap HTML layout and styles

**BREAKING CHANGE**: API changes and dropping support for passing custom icons, alternatively, the custom icon can now be overridden with CSS styles


src/Accordion/Accordion.tsx
```diff
@@ -1,174 +1,98 @@
...
+    React.useEffect(() => constructIds(), [props.list]);
+    React.useEffect(() => setActive(props.defaultExpanded), [props.defaultExpanded]);
+    React.useEffect(() => setId(props.id || randomId("accordion-")), [props.id]);
     React.useEffect(() => {
-        constructIds();
-    }, [props.list]);
-    React.useEffect(() => constructClassName(), [props.className, props.alternative]);
-    React.useEffect(() => constructItemClassName(), [props.iconPosition, props.customIconExpanded, props.iconRotation]);
-    React.useEffect(() => toggle(props.activeIndex), [props.activeIndex]);
+        let cn: string = "seb accordion";
```
kherP, line 42: this classname is meant to be like this? or `seb-accordion`?

yousifalraheem: `accordion` is predefined from bootstrap so I don't want to change it. However, to customize it a little bit and wanted to differentiate it from the rest. Like if you chose to build your own custom accordion, our styles will not have any effects on it. Only components that require a bit of customization over Bootstrap will have `seb` class appended to the main parent.

In short, yes, it is meant to be like this 😄

src/Accordion/Accordion.tsx
```diff
@@ -1,174 +1,98 @@
...
+        cn += props.alternative ? " alternative" : "";
```
omarVodiaK, line 43: I suggest to use something like https://www.npmjs.com/package/classnames or create or own utility for conditionally joining classNames together in frontend-tools

yousifalraheem: implemented it 👍

develop/components/pages/AccordionPage.tsx
```diff
@@ -1,35 +1,16 @@
...
-);
-const angleDoubleRightIcon: JSX.Element = (
-    <svg name="angle-double-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512">
-        <path d="M166.9 264.5l-117.8 116c-4.7 4.7-12.3 4.7-17 0l-7.1-7.1c-4.7-4.7-4.7-12.3 0-17L127.3 256 25.1 155.6c-4.7-4.7-4.7-12.3 0-17l7.1-7.1c4.7-4.7 12.3-4.7 17 0l117.8 116c4.6 4.7 4.6 12.3-.1 17zm128-17l-117.8-116c-4.7-4.7-12.3-4.7-17 0l-7.1 7.1c-4.7 4.7-4.7 12.3 0 17L255.3 256 153.1 356.4c-4.7 4.7-4.7 12.3 0 17l7.1 7.1c4.7 4.7 12.3 4.7 17 0l117.8-116c4.6-4.7 4.6-12.3-.1-17z" />
-    </svg>
-);
+const AccordionPage: React.FC = () => {
+    const [lists] = React.useState<Array<Array<AccrodionListItem>>>(
+        Array(4)
+            .fill(0)
```
DanSallau, line 10: Is Array.fill safe? I am not sure if it browser support is to the right extent or is it ? I had used 

`[ ...Array(4).keys() ].map( () => generateList());`

yousifalraheem: I could do `[0, 0, 0, 0].map()` but it's just easier

yousifalraheem: it works on IE though

PR #291, opened 2020-04-15
refactor(breadcrumb): refactor breadcrumb

yousifalraheem:
refactor breadcrumb to rely heavily on SEB bootstrap styles and layout

**BREAKING CHANGE**: the API for the breadcrumb component has changed


src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,32 +1,46 @@
...
-                            <li key={i} className={"breadcrumb-item" + (i === props.list.length - 1 ? " active" : "")} onClick={(e: React.MouseEvent<HTMLLIElement>) => onClick(i, e)}>
-                                {item}
-                            </li>
-                        ))}
-                    </ol>
-                </nav>
-            </div>
-        );
-    }
+export interface BreadcurmbItem {
```
kherP, line 16: typo `BreadcrumbItem`

kherP: :numario:

develop/components/pages/BreadcrumbPage.tsx
```diff
@@ -1,17 +1,36 @@
...
 import Highlight from "react-highlight";
 const docMD: string = require("../../../src/Breadcrumb/readme.md");
 
-const userIcon: JSX.Element = (
-    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 170 170">
-        <title>regular_black</title>
-        <path d="M149.1,165h-6V132.3c0-18.8-14.2-34.8-32.5-36.8l-24,15.9a3,3,0,0,1-3.3,0L59.2,95.5C41,97.7,26.9,113.6,26.9,132.3V165h-6V132.3c0-22.5,16.7-40.9,38.8-42.9a3,3,0,0,1,1.9.5L85,105.3l23.2-15.4a3,3,0,0,1,1.9-.5c21.9,1.8,39,20.7,39,42.9Z" />
-        <path d="M85,86.4A31.7,31.7,0,0,1,53.4,54.8V36.6a31.6,31.6,0,1,1,63.3,0V54.8A31.7,31.7,0,0,1,85,86.4Zm0-75.5A25.7,25.7,0,0,0,59.4,36.6V54.8a25.6,25.6,0,0,0,51.3,0V36.6A25.7,25.7,0,0,0,85,10.9Z" />
+const houseIcon: JSX.Element = (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
omarVodiaK, line 9: The Viewbox value changed it represents the dimensions of the svg **min-x, min-y, width and height** are you sure you want to change them ?

btw all SEB svgs that i saw so far have this value **viewBox="0 0 170 170"**

yousifalraheem: This is just an SVG that I copied, and the value is totally ignored since I'm overriding it from CSS

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: Great catch!

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,32 +1,46 @@
...
+
+export const Breadcrumb: React.FC<BreadcrumbProps> = React.memo(
+    (props: BreadcrumbProps): React.ReactElement<void> => (
+        <nav aria-label="breadcrumb" className={props.className} id={props.id}>
+            <ol className="seb breadcrumb">
+                {props.list.map((item: BreadcrumbItem, i: number) => {
+                    const isLast: boolean = i === props.list.length - 1;
+
+                    return (
+                        <li key={i} className={"breadcrumb-item" + (isLast ? " active" : "")} aria-current={isLast ? "page" : null}>
```
DanSallau, line 36: Wouldn't this result in a rect key duplicate in an extreme cases ? . I suggest using or appending some variating text to the key.

src/Breadcrumb/readme.md
```diff
@@ -24,14 +24,39 @@ This React component is based on SEB Bootstrap style. Supports customization and
...
 ```
+```typescript
+/** If you use hash router, you need to pass the hash in the href */
+const breadcrumbListObj: Array<BreadcrumbItem> = [
+    { text: "Home", href: "#/" },
+    { text: "Second page", href: "#/second-page" }
+];
+
+function clickHandler(e: React.MouseEvent<HTMLAnchorElement>) {
+    e.preventDefaults();
```
DanSallau, line 35:
```suggestion
    e. preventDefault();
```

develop/components/pages/BreadcrumbPage.tsx
```diff
@@ -1,17 +1,36 @@
...
+const musicIcon: JSX.Element = (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+        <path
+            d="M1088,312a4,4,0,1,1,2-7.445v-14.31l-14,4.49V307a0.925,0.925,0,0,1-.07.332,3.691,3.691,0,0,1,.07.668,4.046,4.046,0,1,1-2-3.445V294a0.9,0.9,0,0,1,1.14-.971l15.62-5.009a0.959,0.959,0,0,1,.2-0.011c0.01,0,.03-0.009.04-0.009a0.146,0.146,0,0,1,.02,0,0.941,0.941,0,0,1,.89.6,0.015,0.015,0,0,1,.01.016c0.01,0.025.03,0.044,0.04,0.07,0.01,0.054,0,.106.01,0.16a0.792,0.792,0,0,1,.03.154v19A4,4,0,0,1,1088,312Zm0-6a2,2,0,1,0,2,2A2,2,0,0,0,1088,306Zm-16,0a2,2,0,1,0,2,2A2,2,0,0,0,1072,306Z"
+            transform="translate(-1068 -288)"
+        />
     </svg>
 );
 
 const BreadcrumbPage: React.FunctionComponent = () => {
```
DanSallau, line 25:
```suggestion
const BreadcrumbPage: React.FunctionComponent = (): React.ReactElement<void> => {
```

yousifalraheem: I don't think it's needed here since we defined as `FunctionComponent` which will prevent you from returning anything other than a `ReactElement`. It's similar to this:
```typescript
const Comp: React.FC = (): boolean => {
    return true;
}
```
Typescript will throw the following error:
```terminal
Error: Type '() => boolean' is not assignable to type 'FC<{}>'.
```

yousifalraheem: I'm only trying to avoid using it because once we start adding these unimportant additions it will screw up the alignment

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: I've read about it and even tried it. The problem is that I cannot create random keys since that would cause the whole list to re-render every time, and I don't have a way of appending the index since all the information passed from the user might be of type `ReactNode`. However, this is not a scenario that raises concerns, simply because this is not a dynamic list where you insert items in between. If the user chooses to do that, React will detect a change in props and cause a render with new keys .

develop/components/pages/BreadcrumbPage.tsx
DanSallau: While it simply adds the return type if the function which is necessary, I noticed we don't have it in either of the demo pages and hence i will agree with you on the structure.

src/Breadcrumb/Breadcrumb.tsx
DanSallau: My concern is not the fact that user can modify the data, my concern is a situation where by user renders multiple instances of the component in the sate UI . In my experience, when in complex or redundant usage, this kind of arrangement tends to throw the duplicate keys errors.

develop/components/pages/BreadcrumbPage.tsx
```diff
@@ -1,17 +1,36 @@
...
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+        <path
+            d="M1088,312a4,4,0,1,1,2-7.445v-14.31l-14,4.49V307a0.925,0.925,0,0,1-.07.332,3.691,3.691,0,0,1,.07.668,4.046,4.046,0,1,1-2-3.445V294a0.9,0.9,0,0,1,1.14-.971l15.62-5.009a0.959,0.959,0,0,1,.2-0.011c0.01,0,.03-0.009.04-0.009a0.146,0.146,0,0,1,.02,0,0.941,0.941,0,0,1,.89.6,0.015,0.015,0,0,1,.01.016c0.01,0.025.03,0.044,0.04,0.07,0.01,0.054,0,.106.01,0.16a0.792,0.792,0,0,1,.03.154v19A4,4,0,0,1,1088,312Zm0-6a2,2,0,1,0,2,2A2,2,0,0,0,1088,306Zm-16,0a2,2,0,1,0,2,2A2,2,0,0,0,1072,306Z"
+            transform="translate(-1068 -288)"
+        />
+    </svg>
+);
+
+const BreadcrumbPage: React.FC = () => {
+    const history: H.History = useHistory();
```
mario-subo, line 26: I recommend importing `RouteComponentProps from react-router` and using `RouteComponentProps["history"]` here so you don't have to import H from history

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: That wouldn't be an issue, the key is only used by the host component

PR #296, opened 2020-04-17
refactor(icon): icon component is removed in favor of svgr loader

yousifalraheem:
**BREAKING CHANGE**: The icon component has been removed in favor of SVGR loader which works much better.


develop/components/pages/BreadcrumbPage.tsx
```diff
@@ -3,25 +3,10 @@ import H from "history";
 import { useHistory } from "react-router";
 import { Breadcrumb, BreadcrumbItem } from "../../../src/Breadcrumb/Breadcrumb";
 import Highlight from "react-highlight";
+import { ReactComponent as HouseIcon } from "../../assets/icons/house.svg";
```
mario-subo, line 6: how come you have to do `{ ReactComponent as HouseIcon }`? Would `import HouseIcon from "../.."` not work?

DanSallau: I think because he defined a generic typing meaning for all .*svg files in the index.d.ts file below.

yousifalraheem: Well, according to [SVGR's docs](https://react-svgr.com/docs/webpack/#usage) it should be just
```typescript
import HouseIcon from "../../assets/icons/house.svg";
```
However, according to my testing, that didn't yield the same result. The default value from the import turned out to be the path to the SVG.

The definition I had in `index.d.ts` [file](https://github.com/sebgroup/react-components/blob/872c820d0b879d4a8d1afb459efaf1ead5ee6276/develop/typings/index.d.ts) shows exactly what the import returns.

yousifalraheem: Honestly, I'm not using it as the way they described in their docs. This is my config in webpack and it's missing typescript
```javascript
{
    test: /\.svg$/,
    include: [path.resolve(__dirname, "develop/assets/svgs")],
    loader: "@svgr/webpack",
    options: { memo: true },
},
```
it should be like this:
```javascript
{
    test: /\.svg$/,
    include: [path.resolve(__dirname, "develop/assets/svgs")],
    loader: "@svgr/webpack",
    options: { memo: true, typescript: true },
},
```
But enabling typescript would fail the build for some reason. Maybe we can improve this in the future

yousifalraheem: Fixed it

PR #307, opened 2020-04-21
Refactor/improve accordion and breadcrumb

yousifalraheem:
Tested on:

- Chrome (Latest)
- Firefox (Latest)
- Safari (Latest)
- Internet Explorer (Latest)


develop/components/pages/AccordionPage.tsx
```diff
@@ -1,15 +1,12 @@
...
+import { AccordionItemProps, AccordionItem } from "../../../src/Accordion/AccordionItem";
 const docMD: string = require("../../../src/Accordion/readme.md");
 
 const AccordionPage: React.FC = () => {
-    const [lists] = React.useState<Array<Array<AccrodionListItem>>>(
-        Array(4)
-            .fill(0)
-            .map(() => generateList())
-    );
+    const [lists] = React.useState<Array<Array<AccordionItemProps>>>([...Array(4)].map(() => generateList()));
```
DanSallau, line 9:
```
    const lists: Array<Array<AccordionItemProps>>> = React.useMemo(() =>[...Array(4)].map(() => generateList()), []);
```
You are not setting it anywhere or are you ? Use memo instead.

mario-subo: Also I'm really not a fan of `Array<Array<T>>` it just looks so ugly. `T[][]`looks much better

develop/components/pages/AccordionPage.tsx
```diff
@@ -39,20 +50,36 @@ const AccordionPage: React.FC = () => {
...
-function generateList(): Array<AccrodionListItem> {
+const Code: React.FC = (props) => <code className="d-inline">{props.children}</code>;
+
+function generateList(): Array<AccordionItemProps> {
     return [
         {
             header: loremIpsum({ units: "words", count: 3 }),
             subHeader: loremIpsum({ units: "words", count: 5 }),
-            content: (
+            children: (
```
mario-subo, line 82: children makes it sound like it should be an array. maybe you should have left it as content 🤣

yousifalraheem: I only removed content, you can render like this:
```tsx
<AccordionItem>Some content</AccordionItem>
```
Which is what the majority of users might prefer. However, if you want to generate a list yourself, you can always use the predefined `children` prop.

develop/components/pages/AccordionPage.tsx
yousifalraheem: @DanSallau 
I only generate once and it's stored in a state, I don't think I need to use memo especially since it's meant to generate different results every time. What's the point of memoizing the results. This is only meant to make it easier for us developers to add more examples by just increasing the number which is currently **4 = 4 examples**

@mario-subo 
We have a standard of using this which is `Array<T>` and it's everywhere. If it comes to preference, I prefer this one because for me it's not about having a short version of it. It's about readability. Using `T[][]` might confuse the reader about more complicated structures such as:
```typescript
type A = Array<[Array<[T]>, K]>; // Current standard
type B = [[T], K][]; // Suggested standard
```

src/Accordion/Accordion.tsx
```diff
@@ -1,95 +1,61 @@
...
+
+export const Accordion: React.FC<AccordionProps> = React.memo(({ list, alternative, onToggle, ...props }: AccordionProps) => {
+    const [active, setActive] = React.useState<number>(Number(props.defaultValue));
+    const [id, setId] = React.useState<string>(props.id);
+    const [className, setClassName] = React.useState<string>("seb accordion");
+
+    React.useEffect(() => {
+        let defaultChecked: number;
+        defaultChecked = list?.findIndex((item: AccordionItemProps) => item.defaultChecked);
+        defaultChecked = React.Children.toArray(props.children).findIndex((Child: React.ReactElement<AccordionItemProps>) => Child.props?.defaultChecked);
```
omarVodiaK, line 27: you changed the value of defaultChecked twice(Line26 and Line 27) without using it elsewhere, i believe the first line is not required anymore?
can you double check.

src/Accordion/Accordion.tsx
```diff
@@ -1,95 +1,61 @@
...
 
     /**
      * Handles accordion item click event
      * @param {React.MouseEvent<HTMLButtonElement>} e MouseEvent
      */
-    const onToggle: React.MouseEventHandler<HTMLButtonElement> = React.useCallback(
+    const onToggleInner: React.MouseEventHandler<HTMLButtonElement> = React.useCallback(
         (e: React.MouseEvent<HTMLButtonElement>) => {
-            setActive(active === Number(e.currentTarget.value) ? null : Number(e.currentTarget.value));
+            setActive(active === Number(e.currentTarget.dataset.indexNumber) ? null : Number(e.currentTarget.dataset.indexNumber));
```
omarVodiaK, line 39:
```suggestion
            setActive(active === Number(e.currentTarget.dataset.indexNumber) ? null : Number(e.currentTarget?.dataset?.indexNumber));
```

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,46 +1,57 @@
...
-    title?: string;
-}
-
-export const Breadcrumb: React.FC<BreadcrumbProps> = React.memo((props: BreadcrumbProps) => (
-    <nav aria-label="breadcrumb" className={props.className} id={props.id}>
-        <ol className="seb breadcrumb">
-            {props.list.map((item: BreadcrumbItem, i: number) => {
-                const isLast: boolean = i === props.list.length - 1;
-                const className: string = classNames(["breadcrumb-item", { active: isLast }]);
+    const isActive: (source: "list" | "children", index: number) => boolean = React.useCallback(
```
omarVodiaK, line 25: is possible to create a type for source ?
you can use it is the switch case line 29

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,46 +1,57 @@
...
+        (source: "list" | "children", index: number): boolean => {
+            const listCount: number = props.list?.length || 0;
+            const childrenCount: number = React.Children.toArray(props.children).length;
+            switch (source) {
+                case "children":
+                    return index === childrenCount - 1;
+                case "list":
+                    return !childrenCount ? index === listCount - 1 : false;
```
omarVodiaK, line 33:
```suggestion
                    return childrenCount ? false : index === listCount - 1;
```

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,46 +1,57 @@
...
+            }
```
omarVodiaK, line 34: i think you should add default case:

src/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,30 @@
...
+        /** Event handler triggered when the links is clicked */
+        onNavigate?: React.MouseEventHandler<HTMLAnchorElement>;
+    }
+>;
+
+export const BreadcrumbItem: React.FC<BreadcrumbItemProps> = React.memo(({ href, onNavigate, ...props }: BreadcrumbItemProps) => {
+    const [active, setActive] = React.useState<boolean>(false);
+    const [className, setClassName] = React.useState<string>("breadcrumb-item");
+
+    React.useEffect(() => setActive(props["data-active"] !== undefined && JSON.parse(props["data-active"])), [props["data-active"]]);
```
omarVodiaK, line 20: JSON.parse("") will throw an error, to avoid i suggest to replace
 ```typescript props["data-active"] !== undefined```
```suggestion
    React.useEffect(() => setActive(props["data-active"] && JSON.parse(props["data-active"])), [props["data-active"]]);
```

src/__utils/Collapse.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export type CollapseProps = React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>;
+
+export const Collapse: React.FC<CollapseProps> = React.memo((props: CollapseProps) => {
+    const collapseRef: React.MutableRefObject<HTMLDivElement> = React.useRef<HTMLDivElement>();
+    const [height, setHeight] = React.useState<number>(0);
+
+    React.useEffect(() => {
+        if (collapseRef && props["data-toggle"] !== undefined) {
+            setHeight(JSON.parse(props["data-toggle"]) ? collapseRef.current.scrollHeight : 0);
```
omarVodiaK, line 11: Its dangerous to parse only after checking for undefined, either
```typescript
            try{
              setHeight(JSON.parse(props["data-toggle"]) ? collapseRef.current.scrollHeight : 0);
            }
            catch(){
              setHeight(0);
            }
```
or add more checks in the condition above

```typescript
if (collapseRef && props["data-toggle"]) {
```

develop/components/pages/AccordionPage.tsx
DanSallau: I think your explanation above is a textbox definition why you need the useMemo. You can memoized it without changing the state. I still believe useMemo is more appropriate here since you don't update your state.

omarVodiaK:
```typescript
  type A = Array<Array<T>, K>;
```

yousifalraheem:
```typescript
let example: A = [
    [
        [1, 2, 3],
        1
    ],
    [
        [1, 2, 3],
        1
    ]
]
```

yousifalraheem:
```typescript
interface ReactChildren {
    map<T, C>(children: C | C[], fn: (child: C, index: number) => T): C extends null | undefined ? C : Array<Exclude<T, boolean | null | undefined>>;
    forEach<C>(children: C | C[], fn: (child: C, index: number) => void): void;
    count(children: any): number;
    only<C>(children: C): C extends any[] ? never : C;
    toArray(children: ReactNode | ReactNode[]): Array<Exclude<ReactNode, boolean | null | undefined>>;
}
```

src/Accordion/Accordion.tsx
yousifalraheem: 🤔

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: added

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: added

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: replaced it with ternary condition, it's more of an if/else kind of relationship

src/__utils/Collapse.tsx
yousifalraheem: updated

src/Breadcrumb/BreadcrumbItem.tsx
yousifalraheem: Fix in a different way

src/Accordion/Accordion.tsx
yousifalraheem: This case will never happen because Accordion injects the "data-index-number" to any element. Even testing it is impossible since there is no case where it doesn't exist.

src/Accordion/Accordion.tsx
yousifalraheem: It is used

src/Accordion/Accordion.tsx
```diff
@@ -1,95 +1,88 @@
...
+         * `active` is not initialized with `props.defaultValue` (`Number(undefined) === NaN`).
+         * Should only run once. The active state will never be set to NaN.
+         *
+         * At least one child is needed to determine the default.
+         */
+        if (list?.length || React.Children.count(props.children)) {
+            // If the default value is set from the parent, it should override anything else
+            let defaultValue: number = -1;
+            // Trying to find an item set to active in the list prop first
+            defaultValue = list ? list.findIndex((item: AccordionItemProps) => item.defaultChecked) : -1;
```
omarVodiaK, line 58:
```suggestion
           // Trying to find an item set to active in the list prop first
           // If the default value is set from the parent, it should override anything else
            let defaultValue: list ? list.findIndex((item: AccordionItemProps) => item.defaultChecked) : -1;
```

src/Accordion/Accordion.tsx
```diff
@@ -1,95 +1,88 @@
...
+            if (defaultValue === -1) {
+                // If the active is not found in one of the list items maybe it's one of the children
+                defaultValue = React.Children.toArray(props.children).findIndex((Child: React.ReactElement<AccordionItemProps>) => Child.props?.defaultChecked);
+                if (list) {
```
omarVodiaK, line 62: i dont think thi is necessery because if defaultValue === -1 it means there is no list since you just checked it in line 58

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,46 +1,58 @@
...
-    <nav aria-label="breadcrumb" className={props.className} id={props.id}>
-        <ol className="seb breadcrumb">
-            {props.list.map((item: BreadcrumbItem, i: number) => {
-                const isLast: boolean = i === props.list.length - 1;
-                const className: string = classNames(["breadcrumb-item", { active: isLast }]);
+    const isActive: (source: ItemsSource, index: number) => boolean = React.useCallback(
+        (source: ItemsSource, index: number): boolean => {
+            const listCount: number = list?.length || 0;
+            const childrenCount: number = React.Children.toArray(props.children).length;
+            if (source === "children") {
```
omarVodiaK, line 31: is there a way to check the type, instead of a harcoded string
```suggestion
            if (source === ItemsSource.children) {
```

src/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+    onNavigate?: React.MouseEventHandler<HTMLAnchorElement>;
+};
+
+export const BreadcrumbItem: React.FC<BreadcrumbItemProps> = React.memo(({ href = "#", onNavigate, ...props }: BreadcrumbItemProps) => {
+    const [className, setClassName] = React.useState<string>("breadcrumb-item");
+
+    React.useEffect(() => setClassName(classnames(["breadcrumb-item", { active: props.defaultChecked }, props.className])), [props.defaultChecked, props.className]);
+
+    return (
+        <li {...props} className={className} aria-current={props.defaultChecked ? props["aria-current"] || "page" : null}>
```
omarVodiaK, line 20: i dont understand why "page"

src/Accordion/Accordion.tsx
mario-subo: correct -> `if (list)` will never be true here.

src/Accordion/Accordion.tsx
```diff
@@ -1,95 +1,88 @@
...
-    /** The content of the accordion item */
-    content: React.ReactNode;
-}
-
-export interface AccordionProps {
-    /** Element class name */
-    className?: string;
-    /** Element id */
-    id?: string;
+export type AccordionProps = React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> & {
```
mario-subo, line 7: Looks like you are including `DetailedHTMLProps` in your accordion props which makes it seem like any of the standard html attributes can be applied to this component which should be mapped accordingly to the html div element but you're not doing that. Or am I missing something? 

For example according to the Typescript definition it is correct for me to do this `<Accordion tabIndex={-1} /> ` but how am I supposed to know that if I pass this value it is never being used?

src/__utils/Collapse.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+export interface CollapseProps extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {
```
mario-subo, line 3: same thing here, check my earlier comment about using `DetailedHTMLProps `

mario-subo:
I will continue with the review of breadcrumbs later. Next time you should separate all changes into separate PRs if they are unrelated. Like breadcrumbs and Accordion.

src/Accordion/readme.md
```diff
@@ -17,49 +17,70 @@ Type: Form Component
...
 
-## Footnote
+#### AccordionProps
+This interface extends all native attributes of `HTMLDivElement`, adding the following extra attributes:
 
-1. Icon positions supported are: `right` and `left`
-2. Icon transitions supported are: `"deg-180"`, `"deg-180-counter"`, `"deg-90"`, `"deg-90-counter"`
-3. Property `list` has an exported interface named `AccordionListItem`:
+| Property     | Type                                         | Description                                                    |
+| ------------ | -------------------------------------------- | -------------------------------------------------------------- |
```
mario-subo, line 72: you should mention defaultValue here since it's actually being used for something

src/__utils/Collapse.tsx
```diff
@@ -0,0 +1,45 @@
...
+    /** True to expand, false to collapse */
+    toggle?: boolean;
+}
+
+export const Collapse: React.FC<CollapseProps> = React.memo(({ toggle = false, ...props }: CollapseProps) => {
+    const collapseRef: React.MutableRefObject<HTMLDivElement> = React.useRef<HTMLDivElement>();
+    const [height, setHeight] = React.useState<"auto" | number>(0);
```
mario-subo, line 10: Optional: I think you can use React.StyleAttributes<HTMLDivElement>["style"]["height"] and ["display"]

src/Breadcrumb/readme.md
```diff
@@ -17,46 +17,69 @@ Type: UI Component
...
 ```
+
+**onNavigate** handler can be passed to the parent as common handler, or passed to each individual child as a unique handler for each item
+
 ```typescript
 /** If you use hash router, you need to pass the hash in the href */
 const breadcrumbListObj: Array<BreadcrumbItem> = [
-    { text: "Home", href: "#/" },
-    { text: "Second page", href: "#/second-page" }
+    { children: "Home", href: "#/" },
```
mario-subo, line 52: again very confusing to have a property called children which makes it sound like it's an array or some kind and then figuring out it's a string.

src/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import classnames from "classnames";
+
+export type BreadcrumbItemProps = React.DetailedHTMLProps<React.LiHTMLAttributes<HTMLLIElement>, HTMLLIElement> & {
```
mario-subo, line 4: Remove the detailed html props and include only the props which are being used in this component

src/__utils/Collapse.tsx
yousifalraheem: if you go down to the return statement, you will find `{...props}` ... I am passing everything to the component

src/Breadcrumb/BreadcrumbItem.tsx
yousifalraheem: it's part of Bootstrap, an accessibility feature [read here](https://getbootstrap.com/docs/4.4/components/breadcrumb/#example)

src/Breadcrumb/Breadcrumb.tsx
yousifalraheem: It's string literal, you can't assign other values than the one specified or Typescript will not build. It's an easier implementation that declaring, exporting and importing enums

src/Accordion/readme.md
yousifalraheem: Even though `defaultValue` is a native attribute, I agree with the fact that I need to point out that it is used to define the value of the breadcrumb if needed

src/Accordion/Accordion.tsx
yousifalraheem: forgot to add `{...props}` to the component. Just added it.

src/Accordion/Accordion.tsx
yousifalraheem: It is necessary, the reason is if there is a list but the index was not found in the list, the `defaultValue` will be `-1` even though the list exists. If it does exists I need to offset the index (`defaultValue`) with the length of the list.
see example
```tsx
<Accordion list={[ { header: "first", children: "Some content" } ]}>
    <AccordionItem defaultChecked header="Second">Some content</AccordionItem>
    <AccordionItem header="Second">Some content</AccordionItem>
</Accordion>
```
In this example ☝️ .. The default value should be `1` which is the index of the checked child `0` + the length of the list which is `1`. If you imagine having 3 rendered accordion items, then `1` would be the second.

src/Breadcrumb/readme.md
yousifalraheem: I'm defaulting back to the native HTML. The preferred way of adding breadcrumbs is this:
```tsx
<Breadcrumb>
    <BreadcrumbItem>Home</BreadcrumbItem>
    <BreadcrumbItem>Other page</BreadcrumbItem>
</Breadcrumb>
```
But if you chose to generate a list, then you can still tap into the `children` prop, just like you do with normal JS/HTML setup

src/Breadcrumb/BreadcrumbItem.tsx
yousifalraheem: I am spreading the `{...props}` .. all detailed HTML props can be used

src/Breadcrumb/readme.md
mario-subo: So in this case the children in a list  can be anything including just a string?

src/Breadcrumb/BreadcrumbItem.tsx
mario-subo: Ok yes I missed that

src/Accordion/Accordion.tsx
mario-subo: We agree you need to check for the possibility of  list existing and the index not being found but if you check the way you structured your code the if (list) will never be true. SO you have to restructure the code a different way. Maybe check the list at the beginning.

src/Breadcrumb/readme.md
yousifalraheem: yes

src/__utils/Collapse.tsx
yousifalraheem: Tried it, didn't work, however, it's unnecessary since this is used internally and I'm only defining the values that I will be using.

PR #314, opened 2020-04-23
Refactor/carousel

yousifalraheem:
### Slide
slide-2

### Fade
fade-2


develop/components/pages/CarouselPage.tsx
```diff
@@ -19,39 +16,77 @@ const CarouselPage: React.FunctionComponent = () => {
...
+                {
+                    children: (
+                        <div className="p-5 bg-primary">
+                            <h1 className="text-light text-center">Start</h1>
+                            <p className="text-light text-center">
+                                <a className="text-light" href="#">
+                                    test
+                                </a>
+                            </p>
+                        </div>
```
kherP, line 53: I know this is example, but maybe you can write this block in a simple function component to reuse it in line `59-66`, `69-77`, `79-86`

src/Carousel/Carousel.test.tsx
```diff
@@ -1,54 +1,170 @@
...
-        await setTimeout(() => {
-            expect(onChange).toBeCalled();
-            mountedWrapper.unmount();
-            done();
-        }, 2000);
+        act(() => {
+            render(<Carousel />, container);
+        });
+        act(() => jest.runOnlyPendingTimers());
+        expect(container.firstElementChild).toBeDefined();
```
kherP, line 29: if this is null, this will still pass the test. 
```suggestion
        expect(container.firstElementChild).not.toBeNull();
```

src/Carousel/Carousel.tsx
```diff
@@ -1,64 +1,140 @@
...
+    transitionStyle?: "slide" | "fade";
+    /** Enables infinite scrolling */
     infinite?: boolean;
-    list: Array<CarouselItem>;
+    /** Shows clickable indicators at the bottom */
+    showIndicators?: boolean;
+    /** A list of accordion items to be rendered */
+    list?: Array<CarouselItemProps>;
+    /** The index of default active slide */
+    defaultValue?: number;
```
kherP, line 23: i think there's better naming for this, defaultValue is a bit confusing

src/Carousel/CarouselIndicators.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        unmountComponentAtNode(container);
+        container.remove();
+        container = null;
+    });
+
+    it("Should render", () => {
+        act(() => {
+            render(<CarouselIndicators />, container);
+        });
+        expect(container.firstElementChild).toBeDefined();
```
kherP, line 25: if it's null, the test will still be successful
```suggestion
        expect(container.firstElementChild).not.toBeNull();
```

src/Carousel/CarouselItem.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        container.remove();
+        container = null;
+    });
+
+    it("Should render", () => {
+        act(() => {
+            render(<CarouselItem />, container);
+        });
+        act(() => jest.runOnlyPendingTimers());
+        expect(container.firstElementChild).toBeDefined();
```
kherP, line 27:
```suggestion
        expect(container.firstElementChild).not.toBeNull();
```

src/Carousel/Carousel.tsx
yousifalraheem: It’s a native HTML attribute, I’m just reusing it. If all components share HTML native attributes we will have more consistent behavior don’t you think?

develop/components/pages/CarouselPage.tsx
yousifalraheem: done

src/Carousel/Carousel.tsx
```diff
@@ -1,64 +1,140 @@
...
+                    switch (newNav) {
+                        case "prev":
+                            newActive = active === 0 ? (isInfinite ? size - 1 : newActive) : active - 1;
+                            break;
+                        case "next":
+                            newActive = active === size - 1 ? (isInfinite ? 0 : newActive) : active + 1;
+                            break;
+                    }
+                }
+                if (newNav !== undefined && newActive !== undefined) {
```
omarVodiaK, line 82: not really important but maybe good in case you have more variables that needs to check for undefined.
it is supported since ie9
```suggestion
                if ([newNav, newActive].every((el: any) => el !== undefined) {
```

src/Carousel/Carousel.tsx
```diff
@@ -1,64 +1,240 @@
...
+                    newNav !== nav && setNav(newNav);
+                    newActive !== active && setActive(newActive);
+                }
+            },
+            [active, infinite, nav, findNewActive]
+        );
+
+        const triggerAutoplay: VoidFunction = React.useCallback((): void => {
+            if (autoplay && !interrupted.current) {
+                setTimeout(() => {
```
mario-subo, line 114: any timeout in the component should be saved and cleared on uinmount in case the component gets destoryed before the autoplayspeed triggers it.

src/Carousel/Carousel.tsx
```diff
@@ -1,64 +1,240 @@
...
+        React.useEffect(() => setSize(list.length + React.Children.count(props.children)), [props.children, list]);
+        /** Sets the default value, if any. Otherwise default to the first item */
+        React.useEffect(() => setActive(props.defaultValue || 0), [props.defaultValue]);
+        /** Set class names */
+        React.useEffect(() => setClassName(classnames("seb", "carousel", { "carousel-fade": transitionStyle === "fade" }, props.className)), [props.className, transitionStyle]);
+        React.useEffect(() => {
+            swipePos && setComingNext(findNewActive(swipePos < 0 ? "next" : "prev"));
+        }, [swipePos]);
+        /** Triggers autoplay if enabled */
+        React.useEffect(() => triggerAutoplay(), [autoplay]);
```
mario-subo, line 194: I noticed the only dependencies here are `autoplay`, however `trigerAutoplay()` function has more dependencies than autoplay alone (like `autoplaySpeed`). I think this will cause an issue where for example `autoplaySpeed` prop is updated and the trigerAutoplay() callback gets updated but this effect will still not run since it only runs when autoplay changes.

webpack.config.js
```diff
@@ -57,6 +57,7 @@ switch (buildType) {
             devServer: {
                 contentBase: "./dist",
                 port: 3000,
+                host: "0.0.0.0",
```
mario-subo, line 60: never understood why this is 0.0.0.0 ?

yousifalraheem: This will allow the webpack-dev-server to appear on network. You would do that if you want to test the UI on a mobile device in the network

src/Carousel/Carousel.tsx
yousifalraheem: added

src/Carousel/Carousel.tsx
yousifalraheem: `triggerAutoplay` is a `callback` and it caches its own values separately. I'm only triggering the function when `autoplay` prop is changed. I tested it, it seems to be working as intended.

mario-subo: Have you tested if you render the component with autoplay = true and autoplaySpeed = 1000 then you change only autoplaySpeed to 5000 right after the initial render. Will the speed actually be 1000 or 5000? From my understanding and what I explained initially it seems like it will be 1000 meaning it will be stale. But I might be wrong

src/Carousel/Carousel.tsx
```diff
@@ -1,64 +1,250 @@
...
+         * Handles navigating to a slide
+         * @param {NavigateTrigger} e Navigation trigger event
+         */
+        const goToSlide = React.useCallback(
+            (e: NavigateTrigger, slideTo?: NavigationDirection): void => {
+                e.cancelable && e.preventDefault();
+                let newActive: number;
+                let newNav: NavigationDirection;
+                const target: EventTarget & HTMLElement = e.target as any;
+                if (["mousedown", "touchstart"].some((val) => val === e.type)) {
```
DanSallau, line 83:
```suggestion
                if (["mousedown", "touchstart"].some((val: string) => val === e.type)) {
```

src/Carousel/Carousel.tsx
yousifalraheem: it will be for the first round and then changes in the second round. Which is ok for now I guess

PR #376, opened 2020-05-19
Refactor Loader

yousifalraheem:
Refactoring loader and add square type

Screen-Recording-2020-05-19-at-2


develop/components/pages/LoaderPage.tsx
```diff
@@ -1,14 +1,24 @@
...
 import Highlight from "react-highlight";
+import { RadioGroup, RadioListModel } from "../../../src/RadioGroup/RadioGroup";
+import { Link } from "react-router-dom";
+import { Dropdown, DropdownItem } from "../../../src/Dropdown/Dropdown";
+import { CheckBox } from "../../../src/CheckBox/CheckBox";
+import classnames from "classnames";
 const docMD: string = require("../../../src/Loader/readme.md");
 
+type Colors = "primary" | "secondary" | "warning" | "danger" | "success" | "white" | "info" | "light" | "dark" | "body" | "muted";
+const lightColors: Array<Colors> = ["white", "secondary", "light"];
```
mario-subo, line 15:
```suggestion
type Color = "primary" | "secondary" | "warning" | "danger" | "success" | "white" | "info" | "light" | "dark" | "body" | "muted";
const lightColors: Array<Color> = ["white", "secondary", "light"];
```

develop/components/pages/LoaderPage.tsx
```diff
@@ -22,46 +32,87 @@ const LoaderPage: React.FC = () => {
...
+    { value: "md", label: "Medium (md)" },
+    { value: "lg", label: "Large (lg)" },
+];
+
+const typeList: Array<RadioListModel<LoaderType>> = [
+    { value: "spinner", label: "Spinner" },
+    { value: "square", label: "Square" },
+];
+
+const colorList: Array<DropdownItem<Colors>> = [
```
mario-subo, line 103:
```suggestion
const colorList: Array<DropdownItem<Color>> = [
```

develop/components/pages/LoaderPage.tsx
```diff
@@ -1,14 +1,24 @@
...
 const docMD: string = require("../../../src/Loader/readme.md");
 
+type Colors = "primary" | "secondary" | "warning" | "danger" | "success" | "white" | "info" | "light" | "dark" | "body" | "muted";
+const lightColors: Array<Colors> = ["white", "secondary", "light"];
+
 const LoaderPage: React.FC = () => {
-    const [textBox1, setTextBox1] = React.useState<string>("Text box group with loader");
+    const [size, setSize] = React.useState<LoaderSize>("md");
+    const [type, setType] = React.useState<LoaderType>("spinner");
+    const [color, setColor] = React.useState<DropdownItem<Colors>>();
```
mario-subo, line 20:
```suggestion
    const [color, setColor] = React.useState<DropdownItem<Color>>();
```

develop/components/pages/LoaderPage.tsx
```diff
@@ -22,46 +32,87 @@ const LoaderPage: React.FC = () => {
                 <div className="info">
                     <h2>Output</h2>
 
-                    <p>Here is a sample tiny example</p>
                     <div className="result">
-                        <SampleContainer height={100} />
-                        <Loader toggle={true} size="sm" />
+                        <SampleContainer color={lightColors.some((c: Colors) => c === color?.value) ? "#636363" : "#efefef"} height={200} />
```
mario-subo, line 36:
```suggestion
                        <SampleContainer color={lightColors.some((c: Color) => c === color?.value) ? "#636363" : "#efefef"} height={200} />
```

src/Loader/loader.scss
```diff
@@ -0,0 +1,106 @@
...
+    &.loader-fullscreen {
+        left: 0;
+        right: 0;
+        bottom: 0;
+        top: 0;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        flex-direction: column;
+        z-index: $zindex-tooltip + 10;
```
mario-subo, line 30: tooltip is already the highest z index value in bootstrap, why the + 10?

src/Loader/loader.scss
```diff
@@ -0,0 +1,106 @@
...
+    }
+    &.loader-sm {
+        @include setLoaderSize(34px);
+    }
+    &.loader-md {
+        @include setLoaderSize(65px);
+    }
+    &.loader-lg {
+        @include setLoaderSize(100px);
+    }
```
mario-subo, line 52: Can we agree to use `rem` where it makes sense? 
The bootstrap one is also using `rem` makes sense to just follow the same.

src/Loader/loader.scss
yousifalraheem: This is a loader, this makes it higher than a tooltip if it is set to `fullscreen`

src/Loader/loader.scss
yousifalraheem: You're right

develop/styles/shared/customizations.scss
```diff
@@ -8,3 +10,17 @@
         }
     }
 }
+
+/* Utility classes */
+.text-xs {
```
omarVodiaK, line 15: i dont think you need this, those variables are available by importing "~@sebgroup/bootstrap/scss/_variables" the _variable file is already importing  "bootstrap-core/variables" line 537

src/Button/Button.tsx
```diff
@@ -14,15 +14,15 @@ export type ButtonProps = JSX.IntrinsicElements["button"] & {
...
-    ({ theme, size, block, className, ...props }: ButtonProps): React.ReactElement<void> => {
-        const [processedClassName, setProcessedClassName] = React.useState<string>("btn btn-primary");
+    ({ theme = "primary", size, block, ...props }: ButtonProps): React.ReactElement<void> => {
+        const [className, setClassName] = React.useState<string>("btn btn-primary");
 
         React.useEffect(() => {
-            setProcessedClassName(classnames(["btn", "btn-" + (theme || "primary"), size && "btn-" + size, { "btn-block": block }, className]));
-        }, [size, theme, block, className]);
+            setClassName(classnames("rc", "btn", `btn-${theme}`, { [`btn-${size}`]: size, "btn-block": block }, props.className));
+        }, [size, theme, block, props.className]);
```
DanSallau, line 22:
```suggestion
        }, [size, theme, block, className]);
```

src/Loader/Loader.tsx
```diff
@@ -1,47 +1,56 @@
...
+                    "loader-cover": cover,
+                    "loader-fullscreen": fullscreen && !cover,
+                    "loader-backdrop": backdrop,
+                    [`loader-${size}`]: size,
+                    [`loader-${type}`]: type,
+                },
+                props.className
+            )
+        );
+    }, [props.className, cover, fullscreen, size, type]);
```
DanSallau, line 42:
```suggestion
    }, [className, cover, fullscreen, size, type]);
```

src/Loader/Loader.tsx
```diff
@@ -1,47 +1,56 @@
...
-                                <div className="seb-loader-circle" />
-                            </div>
-                            <div className="seb-loader-right">
-                                <div className="seb-loader-circle" />
-                            </div>
-                        </div>
-                    </div>
-                </div>
+    return (
+        toggle && (
```
DanSallau, line 45: What will this render when toggle is false?  Perhaps ternary condition is more feasible ```toggle ? ....```

src/Button/Button.tsx
yousifalraheem: Why? I'm setting the value of `className` based on the value of `props.className`

src/Loader/Loader.tsx
yousifalraheem: I tried it, it renders nothing, it's a boolean value and React understands it

src/Loader/Loader.tsx
yousifalraheem: Same here, I don't need `className` instead I need `props.className`

src/Button/Button.tsx
DanSallau: It is confusing though. Since you spread props and select your props by their keys . Wouldn't there be collision in this case ?  Will className still be valid in this component when  remove the line below ?
` const [className, setClassName] = React.useState<string>("btn btn-primary");
`

src/Loader/Loader.tsx
DanSallau: I think it is a bit confusing considering your choice of passing props. Wouldn't `className` still be valid after you remove 

`    const [className, setClassName] = React.useState<string>("");
`
? Can we rename the local className perhaps ?

src/Button/Button.tsx
yousifalraheem: Spreading the props in this case just means that I'm constructing a bunch of variables, the ones I specified and the rest are parked inside a variable called `props`. Everything that I didn't specify is accessed using `props.variable`. That's why `className` doesn't exist because it's inside `props`

PR #392, opened 2020-06-01
feat(table): fixes table sorting

### LINKED ISSUE #390: Table server sorting always sort with descending order
**Describe the bug**
When using server sorting function in table component, the header will not highlight the sorted column and is always returning `DESC` as the sort order.

**To Reproduce**
1. Define table with server sorting function
2. Set one of the header columns to isSorted = true and isSortedDesc = true
3. The header with isSorted = true is not highlighted [NOK]
4. Click the header again
5. Notice that the order direction return is still `DESC` instead of `ASC` [NOK]

**Expected behavior**
Header with isSorted = true should be highlighted. When using server sort, order direction can be Ascending or Descending.

**Screenshots**
https://stackblitz.com/edit/react-ts-yclund

**Additional context**

DanSallau:
Fixes #390

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -1229,9 +1229,9 @@ const Table: React.FunctionComponent<TableProps> = React.memo(
             const updatedColumns: Array<TableHeader> = props.columns.map((column: TableHeader) => {
                 return {
                     ...column,
-                    isSorted: false,
+                    isSorted: column.isSorted || false,
                     canSort: column.canSort !== undefined ? column.canSort : !!props.sortProps ? true : false,
-                    isSortedDesc: false,
+                    isSortedDesc: column.isSortedDesc || false,
                     filters: column.filters || [],
                 };
             });


src/Table/Table.tsx
```diff
@@ -1229,9 +1229,9 @@ const Table: React.FunctionComponent<TableProps> = React.memo(
             const updatedColumns: Array<TableHeader> = props.columns.map((column: TableHeader) => {
                 return {
                     ...column,
-                    isSorted: false,
+                    isSorted: column.isSorted || false,
                     canSort: column.canSort !== undefined ? column.canSort : !!props.sortProps ? true : false,
-                    isSortedDesc: false,
+                    isSortedDesc: column.isSortedDesc || false,
```
yousifalraheem, line 1234: This doesn't seem right.

When columns are changed you shouldn't just return `isSorted` as `true` when the value is `true`. The value being of the `updatedColumns` are not sorted, it would be an invalid flag, wouldn't it?

DanSallau: No, its a server side sorting and hence the data will be sorted by the server. The table will simply renders the data and will not sort the data.

PR #396, opened 2020-06-01
fix(dropdown): fix dropdown error rendered incorrectly

yousifalraheem:
the error message is being rendered outside the component which might cause alignment issues. also
added border color indicating an error

### DIFF
--- a/src/Dropdown/Dropdown.tsx
+++ b/src/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import * as React from "react";
+import React from "react";
 import { randomId } from "@sebgroup/frontend-tools/dist/randomId";
 import "./dropdown-style.scss";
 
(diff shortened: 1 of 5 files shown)


src/Dropdown/Dropdown.tsx
```diff
@@ -361,121 +361,119 @@ const Dropdown: React.FunctionComponent<DropdownProps> = (props: DropdownProps):
...
-                                {(props.clearable || props.multi) && selectedList.length > 0 ? (
-                                    <div className="dropdown-icon-holder" onClick={shouldDisable ? null : handleClickClear}>
-                                        {timesIcon}
-                                    </div>
-                                ) : null}
-                                <div className="dropdown-icon-holder chevron">{chevronDownIcon}</div>
-                            </div>
-                        </>
-                    ) : (
+        <div className={`dropdown custom-dropdown${shouldDisable ? " disabled" : ""}${props.error ? " has-error" : ""}${props.className ? " " + props.className : ""}`} id={props.id}>
```
omarVodiaK, line 364: is there a way to use classNames dep here ?

yousifalraheem: I'm only using it in the refactoring branch, this is branched out of master as a **hotfix**

PR #445, opened 2020-07-21
fix(tooltip): fix tooltip

kherP:
tooltip is rendered in wrong position in long list

### DIFF
--- a/src/__utils/Overlay/Overlay.tsx
+++ b/src/__utils/Overlay/Overlay.tsx
@@ -66,15 +66,18 @@ const Overlay: React.FC<OverlayProps> = React.forwardRef((props: OverlayProps, r
             if (referenceDomRect.bottom < 0 || referenceDomRect.right < 0 || referenceDomRect.left > window.innerWidth || referenceDomRect.top > window.innerHeight) {
                 overlayContentRef?.current?.blur();
             }
-            setPlacementWithCoords(await overlayPositionChecker.getPosition(props.position || "top"));
+            overlayPositionChecker.getPosition(props.position || "top").then((position: ElementPlacementWithCoord) => {
+                setPlacementWithCoords(position);
+            });
         }
     }
 
     /** Get position within view port */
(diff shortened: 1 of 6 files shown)


src/__utils/Overlay/Overlay.tsx
```diff
@@ -59,22 +59,22 @@ const Overlay: React.FC<OverlayProps> = React.forwardRef((props: OverlayProps, r
...
-    function onScroll(ev: Event): void {
+    async function onScroll(ev: Event) {
         const target: HTMLElement = ev.target as HTMLElement;
         if (props.show && target.contains(props.overlayReference())) {
             const referenceDomRect: DOMRect = props.overlayReference().getBoundingClientRect();
             if (referenceDomRect.bottom < 0 || referenceDomRect.right < 0 || referenceDomRect.left > window.innerWidth || referenceDomRect.top > window.innerHeight) {
                 overlayContentRef?.current?.blur();
             }
-            setPlacementWithCoords(overlayPositionChecker.getPosition(props.position || "top"));
+            setPlacementWithCoords(await overlayPositionChecker.getPosition(props.position || "top"));
```
mario-subo, line 69: `OverlayPositionChecker.getPosition()` is not asynchronous and neither are any of the methods that are being called inside so if adding async before this makes it work it probably means the underlying issue lies somewhere else in your code. You're probably calling something before something else or not waiting for something. I suspect adding a setTimeout instead of async/await would also work but again that is also a kind of hack.

mario-subo: Of course I could be wrong and the underlying DOM manipulation APIs are in fact asynchronous and are supposed to be used like this in which case I would love to see some documentation or references to practices which do this.

kherP: i think i should convert `getPosition` to async as well

mario-subo: You should convert getPosition to async if whatever you're doing inside getPosition is in fact asynchronous. That's the main question.

PR #449, opened 2020-07-27
Feature/table-refactor

DanSallau:


src/Table/sections/helperFunctions.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (useSelection) {
+        sum = sum + 1;
+    }
+
+    if (useGroupBy) {
+        sum = sum + 1;
+    }
+    if (useShowActionColumn) {
+        sum = sum + 1;
+    }
```
yousifalraheem, line 22:
```suggestion
    if (useSelection) {
        sum++;
    }

    if (useGroupBy) {
        sum++;
    }

    if (useShowActionColumn) {
        sum++;
    }
```

src/Table/sections/helperFunctions.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * sort array of tabke rows
+ * @param items table rows array
+ * @param columnName the target column name
+ * @param sortDirection the sort direction
+ * @return Array of tableRow
+ */
+export function sortArray(items: Array<TableRow> = [], columnName: string, sortDirection: sortDirectionTypes): Array<TableRow> {
+    const languages: Readonly<Array<string>> = window.navigator?.languages || ["sw", "en"];
```
yousifalraheem, line 35: I don't think you should force `sw` and `en` on the component. You should find a way to get the language from the `navigator` or any other way.

DanSallau: we are getting from navigator "sw" and "en" are just a fallback incase users browser doesn't support one.

PR #450, opened 2020-07-27
refactor(timer): refactored timer to extend native time html element

yousifalraheem:
BREAKING CHANGE: for more information, read the updated documentation


src/Timer/Timer.tsx
```diff
@@ -1,83 +1,63 @@
...
+    const countDown = React.useRef<number>(0);
+    const timer = React.useRef<any>(0);
+
+    /**
+     * Converts the numeric value into two digits string for display purposes
+     * @param {number} value The numberic value to be converted
+     * @returns The two digits string value
+     */
+    const toDoubleDigits = React.useCallback((value: number): string => {
+        return value < 10 ? `0${value}` : String(value);
```
kherP, line 21: `value.toLocaleString('en-US', {minimumIntegerDigits: 2, useGrouping:false})` maybe this is better? at least we do not need to add the 0 ourselves

src/Timer/Timer.tsx
```diff
@@ -1,83 +1,63 @@
...
+
+export type TimerProps = JSX.IntrinsicElements["time"] & {
+    duration?: number;
+    onTimerEnded?: () => void;
+};
+
+export const Timer: React.FC<TimerProps> = React.memo(({ duration, onTimerEnded, ...props }: TimerProps) => {
+    const [time, setTime] = React.useState<string>("00:00");
+    const countDown = React.useRef<number>(0);
+    const timer = React.useRef<any>(0);
```
DanSallau, line 13: Is there any reason while the type here is any ?

src/Timer/Timer.tsx
```diff
@@ -1,83 +1,63 @@
...
+    /** Clears the timer */
+    const clearTimer = React.useCallback((): void => {
+        countDown.current = 0;
+        timer.current && clearInterval(timer.current);
+    }, []);
+
+    /** Kicks off the timer */
+    const startTimer = React.useCallback((): void => {
+        timer.current = setInterval(() => {
+            if (countDown.current--) {
```
DanSallau, line 33: How does this work ? Find it a little confusing . Isn't -- an operation ?

src/Timer/Timer.tsx
yousifalraheem: updated

src/Timer/Timer.tsx
yousifalraheem: The returned type when setting `setInterval` or `setTimeout` is a number. Typescript will complain about that because it thinks it's of type `NodeJS.Timeout` or `NodeJS.Timer` respectively. Using either of them will throw an error during build. Therefore, I always use `any` for intervals and timeouts.

src/Timer/Timer.tsx
yousifalraheem: I wanted to leave a comment explaining it, but then I thought it's straightforward.

It will check if the value of `countDown` is more than zero before decrementing it by 1.

PR #467, opened 2020-09-09
fix(modal): check if event exist before accessing, give ellipsis width

DanSallau:


omarVodiaK (approved):

mario-subo (approved):

src/Modal/Modal.tsx
```diff
@@ -55,7 +55,7 @@ export const Modal: React.FC<ModalProps> = React.memo(({ trapFocus = true, escap
      */
     const escapeKeyListener = React.useCallback(
         (e: KeyboardEvent): void => {
-            if (e.key.toLowerCase() === "escape" && escapeToDismiss) {
+            if (e?.key?.toLowerCase() === "escape" && escapeToDismiss) {
```
yousifalraheem, line 58: I don't think you need to check if the event exist. It will always exist, if not, there is an issue either in the browser or in React itself

yousifalraheem (changes requested):

PR #474, opened 2020-09-21
test(unit-tests): fix unit test config for gatsby

mario-subo:
- Fixes the config for unit tests since move to gatsby
- fixes the build for production (lib components)
- fixed the build for gatsby (docs)


lib/src/ImageCropper/image-cropper-style.scss
```diff
@@ -1,4 +1,4 @@
-@import "../../node_modules/cropperjs/dist/cropper.css";
+@import "../../../node_modules/cropperjs/dist/cropper.css";
```
yousifalraheem, line 1:
```suggestion
@import "~cropperjs/dist/cropper.css";
```

kherP (approved):

src/components/SideMenu.tsx
```diff
@@ -15,16 +15,15 @@ const SIDE_MENU_STORAGE_KEY = "SIDEMENU";
 const components: Array<ComponentsListItem> = comps.sort((a, b) => (a.name > b.name ? 1 : -1));
 
 function getStoredValue(): boolean {
-    const value: string = localStorage.getItem(SIDE_MENU_STORAGE_KEY);
+    const w: Window | null = typeof window !== "undefined" ? window : null;
+    const value: string = w === null ? "true" : w.localStorage.getItem(SIDE_MENU_STORAGE_KEY);
```
yousifalraheem, line 19: I don't like this solution 😐

yousifalraheem: I understand the need for it, but I feel like the [suggested fix from Gatsby](https://www.gatsbyjs.com/docs/debugging-html-builds/#how-to-check-if-window-is-defined) is way cleaner

yousifalraheem: Check #475 for reference

yousifalraheem (approved):

PR #482, opened 2020-10-13
refactor(image-cropper): refactor image cropper, use hooks

DanSallau:
Refactor image cropper, use hooks instead of class component

### DIFF
--- a/src/ImageCropper/ImagePreview.tsx
+++ b/src/ImageCropper/ImagePreview.tsx
@@ -4,43 +4,20 @@ import "./image-preview-style.scss";
 export interface ImagePreviewProps {
     cropResult?: string;
     previewSrc?: string;
-    handleUploadImage: (e: any, cropResult?: string) => void;
+    handleUploadImage: (e: React.ChangeEvent<HTMLInputElement> & React.DragEvent<HTMLInputElement>, cropResult?: string) => void;
     selectButtonText?: string;
     previewClassName?: string;
 }
-export interface ImagePreviewState {
-    cropDataResult: string;
-}
-export class ImagePreview extends React.Component<ImagePreviewProps, ImagePreviewState> {
-    fileInput: HTMLInputElement;
-    constructor(props: ImagePreviewProps) {
-        super(props);
-        this.state = {
-            cropDataResult: "",
-        };
-        this.onFileInputClick = this.onFileInputClick.bind(this);
-    }
 
-    componentDidMount() {
-        this.setState({ cropDataResult: this.props.previewSrc ? this.props.previewSrc : "" });
-    }
+export const ImagePreview: React.FunctionComponent<ImagePreviewProps> = (props: ImagePreviewProps): React.ReactElement<void> => {
+    let fileInput: React.RefObject<HTMLInputElement> = React.createRef();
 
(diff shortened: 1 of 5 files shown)


src/ImageCropper/ImagePreview.tsx
```diff
@@ -60,34 +37,50 @@ export class ImagePreview extends React.Component<ImagePreviewProps, ImagePrevie
...
+        }
+    }, [props.cropResult, props.previewSrc]);
+
+    return (
+        <div className={"profile-image-container " + props.previewClassName}>
+            <div className="profile-image">{cropDataResult ? <img src={cropDataResult} alt="cropped image" id="profile-image" /> : defaultProfileSvg}</div>
+            <input
+                ref={fileInput}
+                type="file"
+                id="fileInput"
```
yousifalraheem, line 71: maybe we should use a random id here

src/ImageCropper/ImagePreview.tsx
```diff
@@ -60,34 +37,50 @@ export class ImagePreview extends React.Component<ImagePreviewProps, ImagePrevie
...
+        if (props.cropResult) {
+            setCropDataResult(props.cropResult);
+        } else if (props.previewSrc) {
+            setCropDataResult(props.previewSrc);
+        }
+    }, [props.cropResult, props.previewSrc]);
+
+    return (
+        <div className={"profile-image-container " + props.previewClassName}>
+            <div className="profile-image">{cropDataResult ? <img src={cropDataResult} alt="cropped image" id="profile-image" /> : defaultProfileSvg}</div>
```
yousifalraheem, line 67: maybe we should use a random id here

src/ImageCropper/ImagePreview.tsx
```diff
@@ -60,34 +37,50 @@ export class ImagePreview extends React.Component<ImagePreviewProps, ImagePrevie
...
+                type="file"
+                id="fileInput"
+                name="profileImage"
+                accept="image/*"
+                onChange={props.handleUploadImage}
+                onClick={(e: React.MouseEvent<HTMLInputElement, MouseEvent>) => {
+                    if (e.target && (e.target as any).value) {
+                        e.target["value"] = null;
+                    }
+                }}
```
yousifalraheem, line 79: You don't need to verify if `e.target` exist. It's like verifying if React as a library is working as intended. Additionally, why are you setting the value from here? That's direct DOM manipulation

src/ImageCropper/ImagePreview.tsx
kherP: added random id

src/ImageCropper/ImagePreview.tsx
kherP: add random id

src/ImageCropper/ImagePreview.tsx
DanSallau: I cannot remember what is it. But last time we added this to address a specific issue. It might be related to drag and drop image, just a guess though. Also, this field is hidden. It isn't shown, we utilized the change event of the field instead.

yousifalraheem:
Closed in favour for #510

PR #497, opened 2020-12-08
refactor(timeline): refactor timeline

kherP:
refactor timeline with spreaded props


lib/src/Timeline/TimelineItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    <div className="title">{title}</div>
+                    <div className="time-wrapper">
+                        <span className="time">{time}</span>
+                    </div>
+                    {!!desc && <div className="desc">{desc}</div>}
+                </div>
+            </div>
+        </div>
+    );
+};
```
yousifalraheem, line 52:
```suggestion
export type TimelineItemProps = JSX.IntrinsicElements["div"] & {
    /** The timeline item header */
    header: React.ReactNode;
    /** The time stamp of the timeline item indicating when it happened */
    time: string;
    /** The timeline item description */
    description?: React.ReactNode;
    direction?: TimelineDirection;
    position?: VerticalTimelineItemPosition | HorizontalTimelineItemPosition;
};

export const TimelineItem: React.FC<TimelineItemProps> = ({ wrapperProps, header, time, description, direction = "vertical", position = "bottom", ...props }: TimelineItemProps) => {
    const ref = React.useRef<HTMLDivElement>();
    const [height, setHeight] = React.useState<number>(ref?.current?.offsetHeight);

    React.useEffect(() => {
        let observer;
        if (direction === "horizontal") {
            setHeight(ref?.current?.clientHeight);
            observer = new ResizeObserver(() => {
                setHeight(ref?.current?.clientHeight);
            });

            //start observing
            observer.observe(ref.current);
        } else {
            observer?.disconnect();
        }
        return () => {
            observer?.disconnect();
        };
    }, [ref, direction]);
    return (
        <div className={classnames("item-holder", position)} {...props} ref={ref} style={direction === "horizontal" ? { marginTop: `${height}px`, ...(props?.style || {}) } : null}>
            <div className={classnames({ [`direction-${position}`]: position })}>
                <div className="title-wrapper">
                    <div className="title">{header}</div>
                    <div className="time-wrapper">
                        <time className="time" dateTime={time}>{time}</time>
                    </div>
                    {!!desc && <div className="desc">{description}</div>}
                </div>
            </div>
        </div>
    );
};
```

lib/src/Timeline/TimelineItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+export type HorizontalTimelineItemPosition = "top" | "bottom";
+export type TimelineDirection = "vertical" | "horizontal";
+
+export type TimelineItemProps = JSX.IntrinsicElements["div"] & {
+    wrapperProps?: JSX.IntrinsicElements["div"];
+    title: React.ReactNode;
+    time: string;
+    desc?: React.ReactNode;
+    direction?: TimelineDirection;
+    position?: VerticalTimelineItemPosition | HorizontalTimelineItemPosition;
```
yousifalraheem, line 15: Since the direction and position is something controlled by the parent, there is no need to expose a prop for it. The best way I found so far is to pass `data-position` from the parent. In this case, it will not be documented and the user won't know it existed since it's only meant for internal use.

You can remove the direction and position props and just pass it from the parent. The other good part about it is that you can use CSS to apply styles directly to it so you don't need to define a class for it.

src/pages/docs/timeline.tsx
```diff
@@ -1,15 +1,31 @@
 import React from "react";
 import Docs from "components/Docs";
-import { Timeline, TimelineListItem } from "@sebgroup/react-components/Timeline";
+import { Timeline, TimelineItemProps } from "@sebgroup/react-components/Timeline";
 import { DynamicFormOption, DynamicFormSection, useDynamicForm } from "hooks/useDynamicForm";
+import TimelineItem from "@sebgroup/react-components/Timeline/TimelineItem";
```
yousifalraheem, line 5:
```suggestion
import { Timeline, TimelineItem, TimelineItemProps } from "@sebgroup/react-components/Timeline";
import { DynamicFormOption, DynamicFormSection, useDynamicForm } from "hooks/useDynamicForm";
```

lib/src/Timeline/TimelineItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+                        <span className="time">{time}</span>
+                    </div>
+                    {!!desc && <div className="desc">{desc}</div>}
+                </div>
+            </div>
+        </div>
+    );
+};
+
+export default TimelineItem;
```
yousifalraheem, line 54:
```suggestion
```

lib/src/Timeline/Timeline.tsx
```diff
@@ -1,119 +1,52 @@
 import * as React from "react";
 import classnames from "classnames";
 import "./timeline-style.scss";
+import TimelineItem, { HorizontalTimelineItemPosition, TimelineDirection, TimelineItemProps, VerticalTimelineItemPosition } from "./TimelineItem";
```
yousifalraheem, line 4:
```suggestion
import { TimelineItem, HorizontalTimelineItemPosition, TimelineDirection, TimelineItemProps, VerticalTimelineItemPosition } from ".";
```

yousifalraheem (changes requested):

yousifalraheem:
This PR requires a lot of clean up. Please do not merge until I clean it up a bit

yousifalraheem (approved):

mario-subo (approved):

PR #498, opened 2020-12-09
Feat/dropdown upgrades

mario-subo:
TODO:
- [x] map old props to new and add deprecation warning
- [x] fix bugs with docs page
- [x] Update unit tests
- [x] remove the deprecated props directly
- [x] Automatically switch to native on mobile

### DIFF
--- a/lib/src/Dropdown/Dropdown.tsx
+++ b/lib/src/Dropdown/Dropdown.tsx
@@ -394,134 +394,137 @@ export const Dropdown: React.FC<DropdownProps> = ({
         <div {...wrapperProps} className={classnames("dropdown custom-dropdown", { disabled: shouldDisable }, { "has-error": error }, wrapperProps?.className)}>
             {label && <label className={classnames("dropdown-label", { disabled: shouldDisable })}>{label}</label>}
 
-            <select
-                {...props}
-                value={value || ""}
-                multiple={multiple}
-                disabled={shouldDisable}
-                className={classnames(`form-control custom-select custom-native-dropdown`, { disabled: shouldDisable }, props?.className)}
-                id={id}
-                hidden={!isMobile()}
-                onChange={(e) => onChange(multiple ? Array.from(e.target.selectedOptions, (option) => option.value) : e.target.value)}
-            >
-                {list.map((item: DropdownItem, i: number) => (
-                    <option hidden={!isMobile()} key={i} value={item.value}>
-                        {item.label}
-                    </option>
-                ))}
-            </select>
+            {isMobile() && (
+                <select
+                    {...props}
+                    value={value || ""}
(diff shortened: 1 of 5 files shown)


lib/src/Dropdown/Dropdown.tsx
```diff
@@ -27,25 +28,33 @@ export interface DropdownPlaceholders {
...
+    onChange: (value: DropdownValue) => void;
     className?: string;
     clearable?: boolean;
     disabled?: boolean;
     error?: string;
-    id?: string;
     label?: string;
     list: Array<DropdownItem>;
     more?: boolean;
+    /** @deprecated use `multiple` instead */
```
kherP, line 44: we have quite a number of components with breaking changes, maybe we can remove this directly?

mario-subo: yeah sounds good. I've added a todo in the PR for it.

lib/src/Dropdown/Dropdown.tsx
```diff
@@ -343,59 +384,51 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
...
                 ref={dropdownToggleRef}
-                className={`btn btn-secondary custom-dropdown-toggle${open ? " open" : ""}${props.more ? " more mx-right" : ""}${shouldDisable ? " disabled" : ""}`}
+                className={classnames(`btn btn-secondary custom-dropdown-toggle`, { open }, { more }, { "mx-right": more }, { disabled: shouldDisable }, props?.className)}
                 id={id}
                 aria-label={`Dropdown toggle: ${getTitleLabel()}`}
                 aria-haspopup={true}
                 aria-expanded={open}
                 tabIndex={shouldDisable ? -1 : 0}
                 onClick={shouldDisable ? null : handleClickToggle}
+                hidden={isMobile()}
```
kherP, line 424: maybe we can make this not to render if it's in mobile? !isMobile() && <div ../>?

lib/src/Dropdown/Dropdown.tsx
```diff
@@ -415,16 +448,17 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
                 onKeyDown={handleKeyDownMenu}
                 tabIndex={0}
                 ref={dropdownMenuRef}
-                className={`dropdown-menu custom-dropdown-menu${open ? " show" : ""}${props.more ? " dropdown-menu-right" : ""}`}
+                className={`dropdown-menu custom-dropdown-menu${open ? " show" : ""}${more ? " dropdown-menu-right" : ""}`}
+                hidden={isMobile()}
```
kherP, line 452: same here

omarVodiaK (approved):

PR #507, opened 2020-12-17
Refactor/radio button

kherP:
make radio button as main component


lib/src/RadioButton/RadioGroup/RadioGroup.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
```
yousifalraheem, line 1: Can you please move this component one level back? we don't need to have another folder called RadioGroup to host it. Just drop it at RadioButton. We still haven't decided how to organize helper components but as of now, this will only create confusion when the users want to import the component directly and not from the root

yousifalraheem (approved):

PR #512, opened 2021-01-06
fix(datepicker): fixed timezone offset bug

mario-subo:
offsets the timezone from the date to fix the bug where the day would be set wrongly around midnight


yousifalraheem (approved):

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -57,7 +57,12 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = ({
 
     const getStringFromDate = React.useCallback(
         (d: Date, monthPicker: boolean): string => {
-            return isValidDate(d) ? d?.toISOString()?.substr(0, monthPicker ? 7 : 10) || "" : "";
+            if (isValidDate(d)) {
+                const tzoffset: number = new Date().getTimezoneOffset() * 60000;
```
omarVodiaK, line 61: are you converting to a different timezone here ?

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -107,7 +103,8 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = ({
 
     const handleChangeCustomDay = (e: React.ChangeEvent<HTMLInputElement>): void => {
         if (!monthPicker) {
-            setCustomDay(Number(e.target.value));
+            const v: number = e.target?.value && !Number.isNaN(e.target?.value) ? Number(e.target.value) : null;
```
omarVodiaK, line 106: i think you should check if e.target.value is a number before using it inside Number.isNan(e.traget.value)
```suggestion
            const v: number = e.target?.value && !Number.isNaN(Number(e.target.value)) ? Number(e.target.value) : null;
```

lib/src/Datepicker/Datepicker.tsx
mario-subo: No I'm adjusting the date in a way which removes any timezone offset

lib/src/Datepicker/Datepicker.tsx
mario-subo: good catch, thanks!! I'll update it

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -69,26 +74,18 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = ({
...
+            if (!max || (max && d <= max)) {
+                return true;
             } else {
-                reject();
+                return false;
             }
-        });
+        } else {
+            return false;
+        }
```
omarVodiaK, line 88:
```javascript
if (!min && !max) {
  return true;
} else if (min && d >= min) {
   if (!max || (max && d <= max)) {
       return true;
   }
}
return false;
```

lib/src/Datepicker/Datepicker.tsx
omarVodiaK: so when a user selects a date would there be any information on what timezone that date has ?

omarVodiaK (approved):

lib/src/Datepicker/Datepicker.tsx
mario-subo: If the user passes a date with timezone info in it, the timezone info will be removed and only the date will be respected.

omarVodiaK: i think it's a bit dangerous to do that, especially if you are adding data in a timezone and the data is saved in a server that have a different timezone

mario-subo: Okay I actually found a better way of achieving this so I updated this part. No timezone modification needed.

PR #513, opened 2021-01-06
Feat/rating

omarVodiaK:
- Remove width/height (not keeping correct aspect ratio breaks the hover state) and replace them with dimension prop
- Remove initialValue prop
- Fix bug when steps is higher than 1
- Add cursor pointer
- change input range style to make the thumb behave in a better way and register click when they happen nearby
-


docs/src/pages/docs/rating.tsx
```diff
@@ -88,15 +83,23 @@ const RatingPage: React.FC = (): React.ReactElement<void> => {
...
+        min={React.ReactText}
+        step={React.ReactText}
+        svgname={string}
+        colors={[string, string]}
+        customSVG={JSX.Element}
+        value={number}
+        onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+            setValue(Number(e.target.value));
+        }}
+    />`;
```
yousifalraheem, line 102: The example should resemble the minimum amount of code needed to get it up and running.
```suggestion
    <Rating value={number} onChange={(e) => setValue(e.target.value)} />`;
```

yousifalraheem (approved):

mario-subo (approved):

PR #516, opened 2021-01-12
refactor(components): add forward ref

kherP:
add forward ref to all components as spreaded ref is not allowed in function component

TODO
------
- [x] forward ref for notification
- [x] forward ref for toggle selector


mario-subo:
We will revisit this PR when we have a stable new version. Marking it as wont fix for now

yousifalraheem:
Technically this is a bug fix, without this PR, the users won't be able to get `ref` from components. Consider adding this BEFORE the major release.

yousifalraheem (approved):
Technically this is a bug fix, without this PR, the users won't be able to get `ref` from components. Consider adding this BEFORE the major release.

PR #527, opened 2021-02-12
Refactor/implement namespace pattern

yousifalraheem:
Hello guys,

I recently stumbled upon an [old PR](https://github.com/facebook/create-react-app/pull/8177) in `create-react-app` on how `React.FC` was removed from the template because it was harming the coding experience rather than improving it. It was always a mystery to me why I couldn't implement namespace pattern with function components, it turns out that `React.FC` was the one preventing it from happening. It could be done, but it would not look as good or clear. Namespace pattern is something like this:

```tsx
<Button.Group>
    <Button>First</Button>
    <Button>Second</Button>
    <Button>Third</Button>
</Button.Group>
```

I did this PR quickly, it took me about 2 hours. All the components' functionalities are working as intended, however, the TS Parser is not working well with this pattern, maybe @kherP can help out to fix it. I highly recommend using the pattern in the next release, implementing this pattern later will cause a lot of breaking changes, since the first release contains breaking changes, it would be more suitable to squeeze this one as well. This pattern adds a great deal of flexibility.

Please do not merge until someone (preferable @kherP) comes up with a fix to TS Parser
- [ ] Fix TS Parser to allow namespace pattern


yousifalraheem:
@kherP @mario-subo @omarVodiaK  can you take a look at this?

mario-subo:
@yousifalraheem we have decided not to prioritise this PR for the next release as we are focusing on the beta for now. After the beta is stable and we release a new version we can revisit this.

yousifalraheem:
@mario-subo .. This PR brings breaking changes, if you don't prioritize them for this release, then you should probably scratch them all together.

I don't think it's wise to introduce yet another set of breaking changes after this release. I was trying to bring these changes since we are already breaking most of the previous pattern, and introducing a new documentation to facilitate the transition.

I recommend choosing either to use this pattern or not, you should not add it later. Hopefully this release will carry the last major breaking changes. No one wants to do yet another refactoring.

PR #529, opened 2021-02-26
fix: fix tslib build and test fail

yousifalraheem:
### Description
This PR fixes a build and test fail due to `tslib`


package.json
```diff
@@ -70,7 +70,7 @@
         "rimraf": "^3.0.2",
         "sass": "^1.32.0",
         "semantic-release": "^17.3.1",
-        "tslib": "^2.0.3"
+        "tslib": "^2.1.0"
```
mario-subo, line 73: In my PR I had to remove this dependency because I had build issues with node 15. Are you sure we need tslib here at all? Seems to work fine without it.

yousifalraheem: Actually `tslib` is required when you want to compile to es5. It brings stuff like spread operators, optional chaining, and more. This PR fixes and issue that was caused by incorrect configurations in `tsconfig`.

I have not tested this PR with node 15 since it's still in its early stages. However, you can try to test with it, feel free to checkout the branch and test.

mario-subo: But the tslib is required in the lib folder not the root folder correct? If so I think tslib will be included with the rollup-plugin-typescript

yousifalraheem: The architectural decision we made back then is to share few dependencies between `lib` and `docs`. The rationale is that they both use a lot of components and they both are supposed to be released with the same versions. For example, both should have the same version of React, Bootstrap, Typescript, and so on. `tslib` is one of these shared packages.

kherP (approved):

PR #533, opened 2021-02-26
Update pipelines

yousifalraheem:
### Description
This PR updates the pipeline adding a missing step to run `npm test`. It also introduces 2 environments of the build using node versions `12` and `14`, the latter being the LTS.


kherP (dismissed):

kherP:
@yousifalraheem, do you mind to include beta branch in the pipeline so we can also release beta version?

yousifalraheem:
If that's the path you've decided on, just create a branch called **beta** and it will be released as beta. Bear in mind, docs will only be published from **master** branch.

kherP:
> If that's the path you've decided on, just create a branch called **beta** and it will be released as beta. Bear in mind, docs will only be published from **master** branch.

i don't think it will trigger the workflow for this repo as both workflows are only listening to master branch. Do you mind to create another PR for these pipeline changes by branching out from master so we can update the workflow for beta branch as well?

yousifalraheem:
This PR depends on #529 to be merged first for the build to pass

yousifalraheem:
@mario-subo Could you please look at this error message and let me know what might be the source of the issue?

```
    Expected: "mesec"
    Received: "month"


    > 195 |         expect(monthElement.querySelectorAll("option").item(0).innerHTML.toLowerCase()).toBe("mesec");
```

kherP (approved):

PR #541, opened 2021-03-17
Doc/support multi version

kherP:
add support for previous version of documentations.

1. only support v3.0.0 and above (this is because some libraries in v2 and below return 404 during npm i)
2. using dropdown temporarily for the version selection (the color is a bit ugly)
3. minimum version supported can be updated in gh_pages.yml line 39

## How does the version dropdown list generated?
1. get_tags job in the actions will retrieve the list using github api, and then filter out the versions older than the minimum version set in line 39.
2. the list will be passed to gh_pages job and be appended to the json file in docs/src/assets/jsons/versions.json

## How do we generate previous versions?
1. based on the list from previous job, we added another job to loop through the list.
2. then we will checkout using the tag, and build the documents accordingly.
3. we also do some checking where if the version is 5.0.0 and above, the build directory will be `docs/public` else the build directory will be `docs`
4. then the job will then copy the build directory to a new sub folder where its name will be the version name.

the flow will need to be: get release tags -> build latest documentation -> build legacy documentations

You can view the example here: https://kherp.github.io/react-components/


omarVodiaK:
It's nice to have access to previous versions of the library, however i don't think you need to list down every single release only the major ones.
```json
"include": [
        { "name": "v5.0.0-beta" },
        { "name": "v4.5.2" },
        { "name": "v3.2.0" },
    ]
```
Not sure if v3 is a good thing to have

kherP:
> It's nice to have access to previous versions of the library, however i don't think you need to list down every single release only the major ones.
> 
> ```json
> "include": [
>         { "name": "v5.0.0-beta" },
>         { "name": "v4.5.2" },
>         { "name": "v3.2.0" },
>     ]
> ```
> 
> Not sure if v3 is a good thing to have
> This is how bootstrap is doing it
> image

alright, i will modify the list to only read from the json file and will omit v3

yousifalraheem:
Maybe try to blend it with its surroundings. Ask Boon Ying about the aesthetics 
```css
background: transparent;
border: none;
padding: 0;
font-weight: bold;
min-height: unset;
```

kherP:
> Maybe try to blend it with its surroundings. Ask Boon Ying about the aesthetics
> 
> ```css-scss
> .rc.custom-dropdown .dropdown-toggle {
>   background: transparent;
>   border: none;
>   padding: 0;
>   font-weight: bold;
>   min-height: unset;
> }
> ```

will try it and ask Boon Ying about the aesthetics 😄

mario-subo (approved):

PR #542, opened 2021-03-18
Feature/move dynamic form to lib

mario-subo:
moved the docs only custom hook to be used as a hook exposed by the library itself instead


kherP (approved):
@mario-subo , if not mistaken, in useDynamicForm, there is console.log which should be removed. i might be wrong, but i remember seeing the log when i test the documentation

mario-subo:
> @mario-subo , if not mistaken, in useDynamicForm, there is console.log which should be removed. i might be wrong, but i remember seeing the log when i test the documentation

Yes, you're right! thanks! I removed it

kherP (approved):

PR #554, opened 2021-03-27
fix(checkbox): fix checkbox pointer

### LINKED ISSUE #519: Table select mode checkbox
**Describe the bug**
The experience of row selection is not optimal 

**To Reproduce**
1. Go to Table
2. Click on Row selection
3. Try to select a row 

**Expected behavior**
When table is in select mode the cursor should change to pointer and the hit are should take up the entire cell

**Screenshots**
Screenshot 2021-01-15 at 2 30 32 PM

kherP:
fix checkbox label pointer

fix #519

### DIFF
--- a/lib/src/Checkbox/checkbox.scss
+++ b/lib/src/Checkbox/checkbox.scss
@@ -21,7 +21,6 @@ $checkbox-padding: $checkbox-padding $checkbox-padding $checkbox-padding $checkb
         border: 1px solid transparent;
         border-radius: 4px;
         transition: border-color 200ms;
-
         > .checkbox-description {
             margin: 0;
             color: $gray-600;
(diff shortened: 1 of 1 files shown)


lib/src/Checkbox/checkbox.scss
```diff
@@ -38,5 +37,8 @@ $checkbox-padding: $checkbox-padding $checkbox-padding $checkbox-padding $checkb
                 color: $gray-300;
             }
         }
+        > .custom-control-label {
```
omarVodiaK, line 40: this will make the cursor pointer even if the checkbox is disabled ?
Vanilla and bootstrap libraries doesn't have this are you sure you want to add it to react lib ?

kherP:
closed this as bootrap library does not have this styling

PR #556, opened 2021-03-29
fix(dynamic-forms): dynamic form should render null check fix

mario-subo:
a bug fixed in should render logic to check if state even exists before checking condition

### DIFF
--- a/lib/src/hooks/useDynamicForm.tsx
+++ b/lib/src/hooks/useDynamicForm.tsx
@@ -8,6 +8,7 @@ import { Dropdown, getValueOfMultipleSelect } from "../Dropdown";
 import { Datepicker } from "../Datepicker";
 import { Stepper } from "../Stepper";
 import { RadioButton, RadioGroup } from "../RadioButton";
+import { isEmpty } from "@sebgroup/frontend-tools/isEmpty";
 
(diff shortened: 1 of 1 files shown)


kherP (approved):

lib/src/hooks/useDynamicForm.tsx
```diff
@@ -80,16 +80,16 @@ export function useDynamicForm(sections: DynamicFormSection[]): [() => JSX.Eleme
...
             const { rulerKey, condition, controlType }: Partial<DynamicFormItem> =
                 sections?.find((item: DynamicFormSection) => item.key === sectionKey)?.items?.find((item: DynamicFormItem) => item.key === itemKey) || {};
             if (controlType === "Hidden") {
                 // Marked as hidden, don't render
                 return false;
             }
-            if (rulerKey) {
-                const rulerState: DynamicFormInternalStateValue = (state[sectionKey] as DynamicFormInternalStateSection)[rulerKey];
+            if (typeof rulerKey === "string" && !!rulerKey.length) {
+                const rulerState: DynamicFormInternalStateValue = !!state[sectionKey] ? (state[sectionKey] as DynamicFormInternalStateSection)[rulerKey] : undefined;
```
omarVodiaK, line 92: i think you should check state before trying to get state[sectionKey] because state is using initialState which is an empty object

mario-subo: done

PR #568, opened 2021-04-05
Fix/dropdown disabled and selected options

mario-subo:
Multi select:

before:

Screenshot 2021-04-05 at 2 56 49 PM

After:

Screenshot 2021-04-05 at 2 57 36 PM

### DIFF
--- a/lib/src/Dropdown/Dropdown.tsx
+++ b/lib/src/Dropdown/Dropdown.tsx
@@ -79,14 +79,32 @@ export const Dropdown: React.FC<DropdownProps> = React.forwardRef(({ wrapperProp
         [isMobile, props.multiple, onMultipleChange]
     );
 
-    const selectAll = (forceValue?: boolean | React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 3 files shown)


lib/src/Dropdown/Dropdown.tsx
```diff
@@ -79,14 +79,32 @@ export const Dropdown: React.FC<DropdownProps> = React.forwardRef(({ wrapperProp
...
+    );
+
+    const isAllSelected = (): boolean => {
+        return Array.from(selectRef.current.options).every((_, i) => {
+            const option = selectRef.current.options.item(i);
+            if (option.disabled) {
+                return true;
+            } else {
+                return option.selected;
+            }
```
omarVodiaK, line 106:
```suggestion
          option.disabled ? true : option.selected
```

omarVodiaK (approved):

lib/src/Dropdown/Dropdown.tsx
kherP: do we need that true? option.disabled || option.selected

PR #569, opened 2021-04-05
Feature/dynamic forms extra props and initial values

mario-subo:


lib/src/hooks/useDynamicForm.tsx
```diff
@@ -308,15 +375,15 @@ const DynamicFormItemComponent: React.FC<{
         }
 
         case "Dropdown": {
-            const { name, onChange, placeholder, value, label } = commonProps;
+            const { name, onChange, placeholder, label, value } = commonProps;
```
kherP, line 378: nice change 😆

lib/src/hooks/useDynamicForm.tsx
```diff
@@ -256,23 +321,24 @@ const DynamicFormItemComponent: React.FC<{
     } = {
         name: props.item?.key || "",
         label: props.item?.label || "",
-        value: (props.state as string) || "",
+        value: props.state as any,
```
kherP, line 324: commonProps value has type of `any`, do we still need `as any` here? maybe can move type for commonProps above?

kherP (approved):

omarVodiaK (approved):

lib/src/hooks/useDynamicForm.tsx
mario-subo: 😂 😂 I try I try hahaha

lib/src/hooks/useDynamicForm.tsx
mario-subo: You're right there's n need as it repeats itself, I did this mostly for readability so that it's obvious this value will be mapped as common prop to all components and can really be any value.

PR #578, opened 2021-04-14
docs(build): build using npm7

kherP:


package.json
```diff
@@ -20,17 +20,16 @@
...
         "ui"
     ],
-    "directories": {
-        "doc": "docs",
-        "lib": "lib"
-    },
+    "workspaces": [
+        "docs",
+        "lib"
+    ],
```
yousifalraheem, line 26: This is not what these are. Workspaces work differently from `directories`. Workspaces treats each entry as an isolated project. `docs` and `lib` share dependencies and they're definitely not isolated

yousifalraheem: You can see that your pipeline failed since it couldn't find `jest`. Jest is shared because they're `directories` of the same project, not isolated `workspaces` which is a setup for monorepos. If you choose to go the `workspaces` route, expect to duplicate all these dependencies in both projects, maintaining them separately, they will end up having different version which might cause issues. I highly recommend that you revert this one back to `directories`

kherP: i believe this idea is no longer true in npm 7.  workspaces are now sharing node_modules on root folder, https://docs.npmjs.com/cli/v7/configuring-npm/package-json#workspaces

the reason the build is failing is node 14 is using npm 6 instead of 7. 

The reason we are upgrading to npm7 is --prefix is no longer supported in npm7, thus we added a temporary fix in the build by forcing all the verify_build to use npm6.

Therefore, to ensure the environments to run properly with proper npm version, here are the options we have:
1. keep forcing all builds to use npm6
2. remove node 14 from verify check, and use 15 for now as Node team will only include npm7 in package 15 onwards

yousifalraheem: npm7 supports workspace, but npm6 doesn't. You're using the same pipeline to build both, that's obviously the reason. Maybe you shouldn't switch to npm7 feature so quickly.  Maybe you should just close this PR and rely on the previous setup 🙂

mario-subo: I'm okay with @kherP suggested option 2 (remove node 14 from verify check, and use 15 for now as Node team will only include npm7 in package 15 onwards nodejs/node#37689). LTS will soon be 15 anyway. I think expecting people who maintain (only us right now) to have the latest node installed until LTS is 15 is not a big thing to ask since we all already have it.

eweseong (approved):

PR #586, opened 2021-04-22
release the build changes to use node 15 and npm 7

mario-subo:


.github/workflows/build-check.yml
```diff
@@ -7,16 +7,16 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
       - name: Setup node
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node }}
-      - name: Install npm@6
-        run: npm i npm@6 # to use workspace
+      - name: Install latest npm
+        run: npm i npm@7
```
kherP, line 19: actually we can remove this task

kherP (approved):

.github/workflows/build-check.yml
mario-subo: oh yeah since it will install 7 anyway, will remove it

kherP (approved):

PR #638, opened 2021-07-27
fix(dropdown): fix dropdown select all checked by default

### LINKED ISSUE #637: Dropdown select all checked by default if it's set to multiple
**Describe the bug**
When the dropdown component is set to multiple by default, the select all checkbox is checked even if the value is empty.

**To Reproduce**
1. Create and define a dropdown element with multiple property
2. click on the component
3. see select all checked

**Expected behavior**
select all should not be checked if value is empty

kherP:
fix #637

### DIFF
--- a/lib/src/Dropdown/Dropdown.tsx
+++ b/lib/src/Dropdown/Dropdown.tsx
@@ -22,8 +22,8 @@ const defaultText: Required<DropdownText> = {
     search: "Search...",
 };
 
-export function getValueOfMultipleSelect(select: HTMLSelectElement): string[] {
-    return Array.from(select.options)
+export function getValueOfMultipleSelect(selectOptions: Array<HTMLOptionElement>): string[] {
+    return Array.from(selectOptions)
         .filter((option) => option.selected)
         .map((option) => option.value);
 }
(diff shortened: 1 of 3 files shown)


lib/src/Dropdown/Dropdown.tsx
```diff
@@ -71,41 +71,39 @@ export const Dropdown: React.FC<DropdownProps> = React.forwardRef(
...
 
-        const isAllSelected = (): boolean => {
-            return Array.from(selectRef.current.options).every((_, i) => {
-                const option: HTMLOptionElement = selectRef.current.options.item(i);
+        const isAllSelected = React.useCallback((): boolean => {
+            return selectRefOptions.every((option: HTMLOptionElement) => {
                 return option.disabled ? true : option.selected;
             });
-        };
+        }, [selectRefOptions]);
```
eweseong, line 106: does all selection includes options that are disabled and not selected?

kherP: yes, selectRefOptions returns every option within the select node

eweseong (approved):

mario-subo (approved):

PR #640, opened 2021-07-27
fix(datepicker): datepicker fires onchange on load

### LINKED ISSUE #631: Datepicker fires onChange on render
**Describe the bug**
Datepicker fires onChange on render

**To Reproduce**
1. add Datepicker to your page
2. add console to onChange of that datepicker
3. see console on page reload/ render

**Expected behavior**
onChange should not fire on render

kherP:
fix #631

### DIFF
--- a/lib/src/Datepicker/Datepicker.tsx
+++ b/lib/src/Datepicker/Datepicker.tsx
@@ -37,11 +37,14 @@ interface UnitNames {
     year: string;
 }
 
+type InputRenderType = "custom" | "date" | "month";
+
 export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwardRef(
     (
         { monthPicker, forceCustom, className, value, min, max, disabled, onChange, localeCode = "en", wrapperProps, customPickerSelectProps, ...props }: DatepickerProps,
         ref: React.ForwardedRef<HTMLInputElement>
     ): React.ReactElement<void> => {
+        const [renderType, setRenderType] = React.useState<InputRenderType>("date");
         const isValidDate = React.useCallback((d: Date): boolean => {
             return !!(d && d instanceof Date && !isNaN(d.getTime()));
         }, []);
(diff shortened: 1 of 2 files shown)


mario-subo (approved):
looks good to me :)

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -139,18 +142,36 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwar
...
+                const day: number = monthPicker ? 1 : customDay;
+                const month: number = customMonth;
+                const year: number = customYear;
+                const dateString: string = `${padNumber(year, true)}-${padNumber(month)}-${padNumber(day)}`;
+                const date: Date = new Date(dateString);
+                const m: number = date.getMonth() + 1;
+                if (date.getFullYear() === year && m === month && date.getDate() === day) {
+                    isDateInRange(date, min, max) ? onChange(date) : onChange(null);
+                } else {
+                    onChange(null);
```
eweseong, line 171: from the code, it seems like `onChange` is still being fired on render here

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -239,6 +260,10 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwar
         const handleOnChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
             const { value: changeEventValue } = e.target;
             const value: Date = new Date(changeEventValue);
+            if (min > value || max < value) {
+                onChange(null);
+                return;
+            }
             onChange(value);
```
eweseong, line 267: is there a requirement to trigger `onChange` with null value? if there isn't maybe we can tap into existing helper method and trigger event only when they are valid?

```
if (isDateInRange(value, min, max)) {
  onChange(value);
}
```

lib/src/Datepicker/Datepicker.tsx
kherP: i was thinking to skip this, but added another commit for this

lib/src/Datepicker/Datepicker.tsx
kherP: updated

lib/src/Datepicker/Datepicker.tsx
```diff
@@ -239,7 +261,11 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwar
         const handleOnChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
             const { value: changeEventValue } = e.target;
             const value: Date = new Date(changeEventValue);
-            onChange(value);
+            if (isDateInRange(value, min, max)) {
+                onChange(value);
+                return;
+            }
+            onChange(null);
```
eweseong, line 268: do we need to trigger change event with `null` value when the data is invalid?

eweseong (approved):

PR #643, opened 2021-08-03
Bug/accordion item toggle

kherP:
allows user to pass function on item level and fire the function in onToggle callback


lib/src/Accordion/Accordion.test.tsx
```diff
@@ -123,4 +123,44 @@ describe("Component: Accordion", () => {
...
+        });
+        const firstButton: HTMLButtonElement = container.querySelectorAll<HTMLButtonElement>("button.btn-link").item(0);
+        act(() => {
+            firstButton.click();
+        });
+        expect(mockFn).toBeCalled();
+        act(() => {
+            firstButton.click();
+        });
+        expect(mockFn).toBeCalled();
```
eweseong, line 164: is the duplicated assertion intended?

kherP: let me remove it haha

eweseong (approved):

PR #650, opened 2021-09-07
feat(dynamic-forms): adds warning messages

mario-subo:


lib/src/hooks/useDynamicForm.tsx
```diff
@@ -91,73 +94,83 @@ type OnChangeFormItem = (item: DynamicFormItem) => OnChangeInput;
...
-            let initialValue: any;
+export type FormRenderFunction = () => JSX.Element;
+export type SetDynamicFormState = React.Dispatch<React.SetStateAction<DynamicFormInternalState>>;
+export type SetDynamicFormErrors = React.Dispatch<React.SetStateAction<DynamicFormErrors>>;
+export type SetDynamicFormWarnings = React.Dispatch<React.SetStateAction<DynamicFormErrors>>;
+type isDynamicFormDirty = boolean;
+export type UseDynamicForm = [FormRenderFunction, DynamicFormInternalState, SetDynamicFormState, SetDynamicFormErrors, SetDynamicFormWarnings, DynamicFormMetaData, isDynamicFormDirty];
+export function useDynamicForm(sections: DynamicFormSection[]): UseDynamicForm {
+    const initialState: DynamicFormInternalState = useMemo(() => {
+        let initialFormState: DynamicFormInternalState = {};
```
eweseong, line 105: notice there's no reference changes, perhaps we can use `const initialFormState`

lib/src/hooks/useDynamicForm.tsx
```diff
@@ -91,73 +94,83 @@ type OnChangeFormItem = (item: DynamicFormItem) => OnChangeInput;
...
+        sections?.map((section) => {
```
omarVodiaK, line 106:
```suggestion
        sections?.map((section: DynamicFormSection) => {
```

lib/src/hooks/useDynamicForm.tsx
```diff
@@ -91,73 +94,83 @@ type OnChangeFormItem = (item: DynamicFormItem) => OnChangeInput;
...
+            initialFormState[section?.key] = {};
+            section.items?.map((item) => {
```
omarVodiaK, line 108:
```suggestion
            section.items?.map((item: DynamicFormItem) => {
```

lib/src/hooks/useDynamicForm.tsx
```diff
@@ -91,73 +94,83 @@ type OnChangeFormItem = (item: DynamicFormItem) => OnChangeInput;
...
 
-            switch (controlType) {
-                case "Dropdown": {
-                    if (multi) {
-                        if (value !== null && value !== undefined && Array.isArray(value) && (value as any[]).every((x) => typeof x === "string")) {
-                            initialValue = value as string[];
+                switch (controlType) {
+                    case "Dropdown": {
+                        if (multi) {
+                            if (value !== null && value !== undefined && Array.isArray(value) && (value as any[]).every((x) => typeof x === "string")) {
```
omarVodiaK, line 115: if value is an empty array this line will return true, if this is the correct use case then i suggest to use:
```suggestion
                            if (!!(value ?? (Array.isArray(value) && (value as any[]).every((x) => typeof x === "string")))) {
```

omarVodiaK (approved):

eweseong (approved):

lib/src/hooks/useDynamicForm.tsx
eweseong: on top of Omar's comment, could we let the `.isArray` to do the checking for `null` and `undefined`?

```
if (Array.isArray(value) && (value as any[]).every((x) => typeof x === "string")) {
```

mario-subo: Omar it's ok to evaluate to true if array is empty as it's a valid state. II will go with Yeoh's comment, I agree the null and undefined check is not needed as will be done is isArray.

PR #656, opened 2021-09-09
fix(datepicker): fix issue where value not set if maxdate is the same date

### LINKED ISSUE #635: Datepicker maxDate validation bug?
**Describe the bug**
When changing maxDate and value via code to higher values that are the same (value and maxDate are the same) everything works okay. When trying to decrease maxDate and value the same way to the same date - validation throws an error and value in GUI stays the same. If value date is at least one day behind maxDate - everything is fine. So it looks like =< is not applied when lowering dates (looks like only < is checked).

**To Reproduce**
Set up Datepicker with maxDate and value as for example 2021-09-09. Then trigger change from somewhere for Datepicker so value and maxDate would be changed to 2021-09-01. It should trigger error. Then try to set the same way maxDate to 2021-09-02 and value to 2021-09-01 - this will not trigger error.

**Expected behavior**
When changing maxDate and value to previous dates at the same time, validation is done according to =<.

mario-subo:
fix #635

### DIFF
--- a/lib/src/Datepicker/Datepicker.tsx
+++ b/lib/src/Datepicker/Datepicker.tsx
@@ -80,6 +80,12 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwar
                 } else {
                     return false;
                 }
+            } else if (max && d <= max) {
+                if (!min || (min && d >= min)) {
+                    return true;
+                } else {
+                    return false;
+                }
             } else {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


lib/src/Datepicker/Datepicker.tsx
```diff
@@ -80,6 +80,12 @@ export const Datepicker: React.FunctionComponent<DatepickerProps> = React.forwar
                 } else {
                     return false;
                 }
+            } else if (max && d <= max) {
+                if (!min || (min && d >= min)) {
+                    return true;
+                } else {
+                    return false;
+                }
```
kherP, line 88:
```suggestion
                return !min || (min && d >= min);
```

kherP (approved):

eweseong (approved):

PR #670, opened 2021-11-24
refactor(stepper): improve stepper accessibility

kherP:
- refactored div to role="spinbutton"
- added more unit tests
- introduced new props accessibilityConfigs to allow user to customize increment/ decrement texts

image


lib/src/Stepper/Stepper.tsx
```diff
@@ -7,52 +7,107 @@ import "./stepper.scss";
...
+                    onDecrement(event);
+                    break;
+                case "ArrowUp":
+                    onIncrement(event);
+                    break;
+                case "ArrowRight":
+                    onIncrement(event);
+                    break;
+            }
+        };
```
eweseong, line 74: should we add `event.preventDefault()` to trap the focus within the stepper component, otherwise clicking on arrow button seemed to move the focus away from the component

eweseong: on a side note, i think we can add the support for `Home` and `End` button pressed per the guideline https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-17

kherP: added 👍

eweseong (approved):

PR #673, opened 2021-11-25
refactor(accordion): improve keyboard support

eweseong:
improve accordion keyboard support and semantics according to [accordion guideline](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html)


lib/src/Accordion/AccordionItem.tsx
```diff
@@ -18,31 +18,32 @@ export const AccordionItem: React.FC<AccordionItemProps> = React.memo(
 
         return (
             <div {...props} ref={ref} className={classnames("rc", "card", { collapsed: !props.defaultChecked }, props.className)}>
-                <div className="card-header" id={uniqueId + "-header"}>
+                <div className="card-header" id={`${uniqueId}--header`}>
```
kherP, line 21: since we are referencing this line in line 40, can we set this to a const and reuse it for easier maintenance

kherP (approved):

lib/src/Accordion/AccordionItem.tsx
eweseong: nice catch, refactored!

PR #676, opened 2021-11-26
refactor(timer): refactor timer

kherP:
- improve accessibility 
- introduce new props timerPrefix and timerSuffix
  - to allow user to pass in prefix/ suffix to be announced along with timer by screen reader


package-lock.json
```diff
@@ -18439,7 +18439,182 @@
                 "treeverse",
                 "validate-npm-package-name",
                 "which",
-                "write-file-atomic"
+                "write-file-atomic",
```
eweseong, line 18442: is the changes a patch from audit?

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
-        this.clearInterval();
-    }
+    const convertToTwoDigits = (rawNumber: number): string => `${rawNumber < 10 ? "0" : ""}${rawNumber}`;
 
-    componentDidUpdate(prevProps: TimerProps): void {
-        if (prevProps.duration !== this.props.duration) {
-            if (this.props.duration !== null && this.props.duration !== undefined) {
-                this.startInterval(this.props.duration);
-            }
+    const convertMStoTime = (milliseconds: number): string => {
```
eweseong, line 45: maybe we can rename this to `formatTime`? generally in date-time related libraries, they reserve `convertXX` for conversion between time units and `formatXyz` for display related transformation

eweseong: or maybe even `formatMilitaryTime`

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
+            clearInterval(innerInterval);
+            setInnerInterval(null);
         }
-    }
+    };
 
-    componentWillUnmount() {
-        this.clearInterval();
-    }
+    const convertToTwoDigits = (rawNumber: number): string => `${rawNumber < 10 ? "0" : ""}${rawNumber}`;
```
eweseong, line 43: maybe we can use `.padStart` for this

```
const formatTwoDigits = (time: number): string => `${time}`.padStart(2, "0");
```

lib/src/Timer/Timer.tsx
eweseong: this algorithm seemed to work as well
```
return new Date(milliseconds).toISOString().slice(11, 19);
```

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
-            if (this.props.duration !== null && this.props.duration !== undefined) {
-                this.startInterval(this.props.duration);
-            }
+    const convertMStoTime = (milliseconds: number): string => {
+        const rawSeconds: number = milliseconds / 1000;
+        const rawMinutes: number = rawSeconds / 60;
+        const displaySeconds: number = rawSeconds % 60;
+        const displayMinutes: number = Math.floor(rawMinutes % 60); // get remainder minutes
+        const displayHours: number = Math.floor(rawMinutes / 60); // get converted hours
+        return (displayHours > 0 ? displayHours + ":" : "") + convertToTwoDigits(displayMinutes) + ":" + convertToTwoDigits(displaySeconds);
```
eweseong, line 51: the following code seemed to work as well
```
return new Date(milliseconds).toISOString().slice(11, 19);
```

kherP: it will not work for duration more than 1 day

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
-        };
-    }
-
-    startInterval(timeout: number): void {
-        this.setState({ timer: this.convertMStoTime(timeout) }, () => {
-            this.clearInterval();
-            this.innerInterval = setInterval(() => {
+    const startInterval = (timeout: number): void => {
+        setTimer(convertMStoTime(timeout));
+        clearInterval();
```
eweseong, line 21: clear interval without interval id seem redundant, is this needed?

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
-                    this.setState({ timer: this.convertMStoTime(timeout) }, () => {
-                        if (timeout === 0) {
-                            this.props.callback();
-                            this.clearInterval();
-                        }
-                    });
+                    setTimer(convertMStoTime(timeout));
+                    if (timeout === 0) {
+                        callback?.();
+                        clearInterval();
```
eweseong, line 29: clear interval without interval id seem redundant, is this needed?

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
+        const displaySeconds: number = rawSeconds % 60;
+        const displayMinutes: number = Math.floor(rawMinutes % 60); // get remainder minutes
+        const displayHours: number = Math.floor(rawMinutes / 60); // get converted hours
+        return (displayHours > 0 ? displayHours + ":" : "") + convertToTwoDigits(displayMinutes) + ":" + convertToTwoDigits(displaySeconds);
+    };
+
+    React.useEffect(() => {
+        if (duration >= 1000) {
+            // only start timer if there's at least 1 second set
+            clearInnerInterval();
```
eweseong, line 57: is this step needed? am assuming on re-render/unmount the clean up code would be called?

package-lock.json
kherP: revert changes

lib/src/Timer/Timer.tsx
kherP: rename function

lib/src/Timer/Timer.tsx
```diff
@@ -6,80 +6,67 @@ export type TimerProps = JSX.IntrinsicElements["div"] & {
...
-    componentDidMount() {
-        if (this.props.duration !== null && this.props.duration !== undefined) {
-            this.startInterval(this.props.duration);
+    const clearInnerInterval = (): void => {
+        if (innerInterval) {
+            clearInterval(innerInterval);
+            setInnerInterval(null);
         }
-    }
+    };
```
eweseong, line 41: tried to debug the timer and notice the interval is still running after the duration has ended, could not pin point the root cause of it; but was able to clear the intervals after merging all the helper functions into the side effects
```
    React.useEffect(() => {
        let intervalId;

        if (duration >= 1000) {
            // only start timer if there's at least 1 second set
            setTimer(convertMStoTime(duration));
            intervalId = setInterval(() => {
                if (duration > 0) {
                    duration -= 1000;
                    setTimer(convertMStoTime(duration));
                } else {
                    callback?.();
                    clearInterval(intervalId);
                }
            }, 1000);
            setInnerInterval(intervalId);
        }

        return () => clearInterval(intervalId);
    }, [duration]);
```

kherP: yes :(  i cleaned up the code

lib/src/Timer/Timer.tsx
kherP: should be clearInnerInterval

eweseong (approved):

eweseong (approved):

PR #678, opened 2021-11-30
Refactor/tab

kherP:
- added keyboard support based on [W3C guidelines](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html)
- introduced new properties:
  - autoSelectOnFocus: set focused element to active by auto activation
  - onTabDelete: to allow user to do some actions on delete button press
- refactored playground component to use tab component

### DIFF
--- a/lib/src/Tabs/Tabs.tsx
+++ b/lib/src/Tabs/Tabs.tsx
@@ -1,22 +1,62 @@
 import React from "react";
 import { TabItemProps } from "./TabItem";
 import classnames from "classnames";
+import { useCombinedRefs } from "../hooks/useCombinedRef";
 
 export type TabsProps = JSX.IntrinsicElements["ul"] & {
     /** Index of focsued tab */
     value?: number;
+    /** set focused index to active index on focus */
+    autoSelectOnFocus?: boolean;
     /** Callback on tab item changed */
     onTabChange?: (index: number) => void;
+    /** Callback on tab item when delete key is triggered */
+    onTabDelete?: (index: number) => void;
 };
 /** Tabs organize and allow navigation between groups of content that are related and at the same level of hierarchy. */
-export const Tabs: React.FC<TabsProps> = React.forwardRef(({ value, onTabChange, ...props }: TabsProps, ref: React.ForwardedRef<HTMLUListElement>) => {
+export const Tabs: React.FC<TabsProps> = React.forwardRef(({ value, onTabChange, autoSelectOnFocus, onTabDelete, ...props }: TabsProps, ref: React.ForwardedRef<HTMLUListElement>) => {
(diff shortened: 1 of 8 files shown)


lib/src/Tabs/Tabs.tsx
```diff
@@ -1,22 +1,62 @@
...
+        }
+    };
     const onClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
-        event.currentTarget.href.endsWith("#") && event.preventDefault();
-        onTabChange && onTabChange(parseInt(event.currentTarget.dataset.indexNumber));
+        navigateToTab(event.currentTarget, event);
+    };
+
+    const onKeyDown = (event: React.KeyboardEvent<HTMLUListElement>) => {
+        const items: HTMLAnchorElement[] = Array.from(tabsRef.current.querySelectorAll(".nav-tabs > .rc.nav-item > .nav-link"));
```
eweseong, line 34: let's move this closer to its usage in the `ArrowX` case to reduce HTML queries

lib/src/Tabs/Tabs.tsx
```diff
@@ -1,22 +1,62 @@
...
+        switch (event.key) {
+            case "Enter":
+            case " ":
+                navigateToTab(event.target as HTMLAnchorElement, event);
+                break;
+            case "ArrowLeft":
+            case "ArrowRight":
+                const direction: number = event.key === "ArrowLeft" ? -1 : 1;
+                const nextFocusedIndex: number = focusedIndex + direction;
```
eweseong, line 43: we can modify the algorithm slightly to support infinite moving (move to last when on first; move to first when on last)

```
const nextFocusedIndex: number = (focusedIndex + items.length + direction) % items.length;
```

lib/src/Tabs/Tabs.tsx
```diff
@@ -1,22 +1,62 @@
...
+    };
     const onClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
-        event.currentTarget.href.endsWith("#") && event.preventDefault();
-        onTabChange && onTabChange(parseInt(event.currentTarget.dataset.indexNumber));
+        navigateToTab(event.currentTarget, event);
+    };
+
+    const onKeyDown = (event: React.KeyboardEvent<HTMLUListElement>) => {
+        const items: HTMLAnchorElement[] = Array.from(tabsRef.current.querySelectorAll(".nav-tabs > .rc.nav-item > .nav-link"));
+        switch (event.key) {
```
eweseong, line 35: maybe we can add support for `Home` and `End` button navigation here as well?

mario-subo (approved):

eweseong (approved):

PR #679, opened 2021-11-30
refactor(chip): improve accessibility

eweseong:
- there is no standard guideline for `chip` component, hence the changes made are referencing from prominent component library like Material UI
- add keyboard support reference from https://mui.com/components/chips/#accessibility; when the `chip` is focused, pressing on `Backspace` or `Delete` key would trigger the on close event.

### DIFF
--- a/lib/src/Chip/Chip.tsx
+++ b/lib/src/Chip/Chip.tsx
@@ -1,15 +1,32 @@
-import React from "react";
 import classnames from "classnames";
+import React from "react";
 import "./chip.scss";
 
 export type ChipProps = JSX.IntrinsicElements["div"] & {
     /** Event triggered when the close button is clicked */
-    onClose: React.MouseEventHandler<HTMLButtonElement>;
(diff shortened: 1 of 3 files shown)


lib/src/Chip/Chip.tsx
```diff
@@ -1,15 +1,32 @@
...
+                    onClose(event);
+                    break;
+                }
+            }
+        },
+        [onClose]
+    );
+
+    return (
+        <div {...props} ref={ref} role="button" className={classnames("rc", "chip", props.className)} onKeyDown={onKeyDown} tabIndex={0}>
```
mario-subo, line 25: Maybe can look into this https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#accessibility_concerns

eweseong: I was worried that changing the wrapper to button would be weird as one of the children is also a button, is it okay to nest button within button semantically?

kherP (approved):

PR #687, opened 2021-12-13
refactor(pagination): improve pagination accessibility

kherP:
introduce new props to support pagination accessibility

https://www.a11ymatters.com/pattern/pagination/


eweseong (approved):

eweseong:
noticed we are not defaulting the recommended `role` and `aria-current` attributes suggested in the guide shared above, was wondering if we are delegating that responsibility to the component's users?
image

eweseong (approved):

PR #691, opened 2021-12-17
refactor(modal): restructure keyboard support

eweseong:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

- restructure keyboard support to rely on react keyboard event over native keyboard event
- replace explicit focus trap element with programmatical approach

### DIFF
--- a/docs/src/pages/docs/modal.tsx
+++ b/docs/src/pages/docs/modal.tsx
@@ -36,6 +36,7 @@ const NotificationPage: React.FC = () => {
                 { key: "position", label: "Position", controlType: "Radio", options: positions, value: positions[0].value },
                 { key: "centered", label: "centered", controlType: "Checkbox", value: false },
                 { key: "fullscreen", label: "fullscreen", controlType: "Checkbox", value: false },
+                { key: "trapfocus", label: "trapfocus", description: "Deprecated. Trap focus would be enabled by default.", controlType: "Checkbox", value: true },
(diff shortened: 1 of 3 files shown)


lib/src/Modal/Modal.tsx
```diff
@@ -23,21 +24,45 @@ export type ModalProps = JSX.IntrinsicElements["div"] & {
     position?: ModalPosition;
     /** Modal toggle */
     toggle?: boolean;
-    /** To only allow tab to focus within modal */
-    trapFocus?: boolean;
```
kherP, line 27: before removing it entirely, should we just mark it @deprecated and keep it until next major release?

docs/src/pages/docs/modal.tsx
```diff
@@ -26,6 +26,7 @@ const positions: Array<DynamicFormOption<ModalPosition>> = [
 
 const NotificationPage: React.FC = () => {
     const [toggle, setToggle] = React.useState<boolean>(false);
+    const toggleButtonRef = React.useRef<HTMLButtonElement>();
```
mario-subo, line 29: missing type

lib/src/Modal/Modal.tsx
eweseong: if I've added in a breaking change commit 57d231c1421c00e4b4946f2322d7bd7530a92e76 while working on one of the accessibility changes, would this bump our major version while preping the next release?

eweseong: after discussion with the team, the `trapFocus` prop would be deprecated regardless if the major version were bumped

kherP (approved):

mario-subo (approved):

PR #756, opened 2022-08-10
Update rollup.config.js

AaronTee:
SSR environment doesn't have window context, therefore referring document explicitly would throw 'ReferenceError: document is not defined.'. Suggest to check by type instead.


sjkayle (approved):

mario-subo (approved):

mario-subo:
included this change in re #760 Thank you!

PR #791, opened 2023-02-02
fix: date picker keyboard error

eweseong:
fixes issue #792 

also added minor refactoring on:
- moving pure functions out of components to reduce the flooding of `React.useCallback()`
- simplify range check algorithm
- eslint setup


mario-subo (approved):

mario-subo:
looks good! thanks for the fix

sjkayle (approved):

